PR #2, opened 2016-11-17
Support for refreshing the list of measurements every 3 hours

hellais:
* Add celery based task runner containers


hellais:
Let's deploy this baby!

PR #11, opened 2016-12-30
Add production environment

anadahz:
* Add Makefile target for production
* Add support for letsencrypt certificates with https-portal
* Add sslmode to example secrets file
* Fix minor link typo in README


docker-compose-production.yml
```diff
@@ -0,0 +1,40 @@
...
+  links:
+    - redis
+    - rabbitmq
+https-portal:
+  image: steveltn/https-portal:1.0.0
+  ports:
+    - "80:80"
+    - "443:443"
+  links:
+    - web
```
hellais: It seems like this is the only thing that is being added to the production docker-compose file. Can you not just add this to `config/production.yml`?

Makefile
```diff
@@ -62,4 +63,8 @@ push-staging:
 	make APP_ENV=staging build
 	docker-compose -f docker-compose.yml -f config/staging.yml up -d
 
-.PHONY: default build serve clean debug develop develop-rebuild dropdb test
+production: APP_ENV=production
+production: DOCKER_COMPOSE_FILE=docker-compose-production.yml
+production: serve-d
```
hellais: Do you need to override this? Is it not possible to just override the `$(APP_ENV).yml` file?

docker-compose-production.yml
anadahz: Apart from this there are a number of things introduced in `docker-compose-production.yml`:
- Removal of the db container
- Addition of the https-portal

@hellais Do you think that we can get these changes in `config/production.yml` ?

Makefile
anadahz: I used the same definitions as in [staging](https://github.com/TheTorProject/ooni-measurements/blob/master/Makefile#L58-L59) and [develop] (https://github.com/TheTorProject/ooni-measurements/blob/master/Makefile#L52-L53) build targets.

docker-compose-production.yml
hellais: So apart from this there is the removal of the db container, correct?

I would see if there is a way to do that in the inherited config file if not then you should move the db container to config files of development, testing, staging and not add it to the production one.

Makefile
hellais: Sorry I commented on the wrong line. What I meant was the overriding of the `DOCKER_COMPOSE_FILE`. We already have splitting of that based on the target definition adding an extra docker file for production makes this not so clean.

docker-compose-production.yml
anadahz: @hellais Did you find a way to do that in the config file without calling a different compose file?

hellais: As I said, if you don't find a way to do that, just rip the db: container definition out of the base docker-compose file and place it inside of dev, testing, staging and do not put it in production.

In production put the https-portal definition and voila, done, no need for a docker-compose-production.

Pasting what I wrote also on IRC for futher context.
```
09:28 < hellais> docker-compose.yml is the base docker compose file (this is shared by EVERY deployment target)
09:29 < hellais> config/$TARGET.yml is the overriden docker compose file that contains changes specific to that target. You can put in here ANYTHING,
                 including the creation of extra containers, linking to other containers, setting up port mapping, etc.
09:30 < hellais> config/$TARGET.env is the environment file definition for a specific target. This contains ONLY environment variable definitions
09:30 < hellais> config/base.env contains the environment variables shared by every target.
```

hellais:
@anadahz I implemented what I was describing in the above thread in https://github.com/TheTorProject/ooni-measurements/commit/072c2eda1e250cd9c5389e5d2d362634857cdf11

config/production.yml
```diff
@@ -3,16 +3,34 @@ web:
...
   - ./config/base.env
   - ./config/production.env
   - /opt/secrets.env
+  links:
+    - redis
+    - rabbitmq
+
 https-portal:
-  env_file:
-  - ./config/production.env
```
anadahz: This is needed by the https-portal to setup and renew the tls certificates.

anadahz:
Let's get this merged.

PR #17, opened 2017-02-22
Fix duplicate DB entries, equal staging/production

anadahz:
* Set redis workers concurrency to 1 to eliminate duplicate DB entries
* Staging should as close as possible to the production build to
  detect issues caused by different software versions and configurations

### DIFF
--- a/config/development.yml
+++ b/config/development.yml
@@ -1,5 +1,5 @@
 db:
-  image: postgres:9.5
+  image: postgres:9.6
   ports:
     - "5433:5432"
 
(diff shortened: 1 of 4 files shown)


hellais:
I think setting the celery worker concurrency to 1 is not the correct fix in this case. The problem was actually that the schedule was not being defined properly and by chance it was working on staging because there the workers are only 1 (due to it having 1 core).

I properly address this issue in https://github.com/TheTorProject/ooni-measurements/pull/18 and added the pins you suggest in there for the DB version and MQs

anadahz:
Closing as this has been addressed in #18.

PR #18, opened 2017-02-22
Fix crontab scheduling to execute every 6 hours

hellais:


anadahz:
I will suggest to set the celery worker concurrency otherwise we can't reproduce the duplicate DB entries bug in any deployment environment with 1 CPU core.

bassosimone:
@anadahz wouldn't it be more robust to have a testing environment with 2 cores such that we can mimic more closely what happens in production? (provided we have enough cores...)

anadahz:
@bassosimone indeed it will be better to have a staging environment with 2 cores (as in production).

I did some further testing and it seems that this PR solve the duplicate reports bug, LGTM.

PR #22, opened 2017-06-17
Feature/pipeline integration

hellais:
This feature branch aims to make the measurements API be backed by the new data processing pipeline.

Next steps:

* [x] Reach feature parity with current measurements API
* [x] Add API endpoint for searching and filtering measurements
* [x] Add API endpoint for retrieving an individual measurement

Some things to remember for deployment:

* [x] Create a special Read-only user for the pipeline measurements table
* [x] Create indexes as I created them on my local copy (see: indexes section)

In order to get decent performances, in particular for mapping `report_id, input`, which is going to be the most common filtering query, I had to create the following indexes:

```
CREATE INDEX measurement_report_no_idx ON public.measurement (report_no);
CREATE INDEX measurement_input_no_idx ON public.measurement (input_no);
CREATE INDEX measurement_id_idx ON public.measurement (id);
CREATE INDEX autoclaved_no_idx ON public.report (autoclaved_no);
```

No surprisingly, query time went from 20+ seconds to < 0.05 seconds.


hellais:
@darkk on the topic of:

> Add API endpoint for retrieving an individual measurement

What I am thinking of is that we probably are better of exploiting data locality by having a minimalistic API that is directly on datacollector (or wherever the actual measurements live) that we can slap a caching proxy in front of even.

This API endpoint would take as input: `frame_off`, `frame_size`, `intra_off`, `intra_size` and `autoclaved_no` (or any other identifier that is kept in the database and can be used to identify the needed lz4 file).
Based on the above it would then return the decompressed payload of the required measurement in the response body and take care of retrieving the autoclaved file, seeking to correct offset and decompressing it.

I am a bit torn on whether or not the decompression should be done by this API of if the measurements backend should do it. I think that given the fact measurements already does a bunch of things and maybe we want to save CPU load there it should be fine to have datacollector or some machine close to it do it.

darkk:
- [ ] unify LZ4 libraries

measurements/config.py
```diff
@@ -22,6 +18,9 @@
 BASE_URL = os.environ.get("BASE_URL",
                           "https://measurements.ooni.torproject.org/")
 
+AUTOCLAVED_BASE_URL = os.environ.get("AUTOCLAVED_BASED_URL",
```
hellais: Fix this environment variable.

hellais:
I am merging this as we have this currently deployed.

Issue #24, opened 2017-07-17
Cleanup after migration to new pipeline


darkk:
The questions are based on feature/pipeline-integration branch, commit 5677f00d781181b1285178cd3314ba8c1cf20ae7

- [x] are WEBSERVER_PORT and WEBSERVER_HOST variables used or gunicorn's `--bind` is used instead? same question for `WORKERS`. 
- [x] should nodish APP_ENV set pythonic DEBUG as well?
- [x] is `PRODUCTION` ever used?
- [x] code that works with S3 seems to be dead, should it be dropped?
- [x] `DATABASE_URL` contains password but no `SECRET` substring, so it is logged, should it be?
- [x] `AUTOCLAVED_BASED_URL` looks like typo
- [x] every single worker tries to update database schema, it produces several errors on startup:

```
2017-07-17T18:31:25.413144442Z sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "pg_type_typname_nsp_index"
2017-07-17T18:31:25.413150845Z DETAIL:  Key (typname, typnamespace)=(report_files_id_seq, 2200) already exists.
2017-07-17T18:31:25.413156771Z  [SQL: '\nCREATE TABLE report_files (\n\tid SERIAL NOT NULL, \n\tprobe_asn VARCHAR(200), \n\tprobe_cc VARCHAR(2), \n\treport_id VARCHAR(2000), \n\ttest_start_time TIMESTAMP WITHOUT TIME ZONE, \n\ttest_name VARCHAR(2000), \n\tidx NUMERIC, \n\tbucket_date VARCHAR(200), \n\tfilename VARCHAR(2000), \n\tPRIMARY KEY (id)\n)\n\n']
```
- [x] `report_files` table was created

hellais:
I did the following:

* Set app["DEBUG"] when APP_ENV == "development"
* Removed dead variables (WEBSERVER_PORT, WEBSERVER_HOST, PRODUCTION, WORKERS)
* Deleted all s3 related dead code (and added some consistency checks to ensure the REPORTS_DIR contains the s3:// prefix)
* Disabled logging of DATABASE_URL
* Fixed the typos
* Removed code for creation of tables
* Remove the reports_files schema

I separated each change in it's own commit to facilitate auditing.

I also seem to remember there was some other bugs in the beta measurements API reported on IRC maybe. Do you recall what that was?

hellais:
[...]

PR #27, opened 2017-08-17
Fix/pipeline integration darkk bugs

hellais:
This implements the feedback @darkk gave in issue #24


Makefile
```diff
@@ -59,8 +62,8 @@ dropdb:
 develop: APP_ENV=development
 develop: .state/docker-build-$(APP_ENV) serve
 
-develop-debug: APP_ENV=development
-develop-debug: .state/docker-build-$(APP_ENV) debug
+develop-shell: APP_ENV=development
+develop-shell: .state/docker-build-$(APP_ENV) shell
```
bassosimone: I never saw this syntax for Makefiles. Does it mean that `APP_ENV=development` is only set (or overriden) for the `develop-shell` target?

config/base.env
```diff
@@ -1,2 +1 @@
 DATABASE_URL=postgresql://postgres@db/measurements
-WEBSERVER_PORT=8000
```
bassosimone: Removed a dead variable.

config/development.env
```diff
@@ -1,4 +1,3 @@
 APP_ENV=development
 REPORTS_DIR=/data/reports/
 BASE_URL=http://localhost:3000/
-WEBSERVER_PORT=3000
```
bassosimone: Removed a dead variable

config/production.env
```diff
@@ -3,4 +3,3 @@ REPORTS_DIR=s3://ooni-public/sanitised
 BASE_URL=https://measurements.ooni.torproject.org/
 # https-portal environment variables
 DOMAINS=measurements.ooni.torproject.org -> http://web:8000, measurements.ooni.io -> http://web:8000
-PRODUCTION=TRUE
```
bassosimone: Removed a dead variable.

bassosimone:
Okay, I am looking into this!

measurements/filestore.py
```diff
@@ -3,124 +3,10 @@
...
-    if app.config['REPORTS_DIR'].startswith("s3://"):
-        return get_download_url_s3(app, bucket_date, filename)
-    return urljoin(
-        app.config['BASE_URL'],
-        '/files/download/{}'.format(filename)
-    )
+    url = "https://s3.amazonaws.com/"
+    # strip the leading s3://
+    url += app.config['REPORTS_DIR'][5:]
+    return urljoin(url, bucket_date, filename)
```
bassosimone: Deleted all s3 related code

measurements/pages/__init__.py
```diff
@@ -10,8 +10,7 @@
 from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 from werkzeug.exceptions import BadRequest, NotFound, HTTPException
 
-from measurements.filestore import get_download_url, gen_file_chunks, \
-    FileNotFound, S3NotConfigured
+from measurements.filestore import get_download_url
```
bassosimone: Do not fetch now-deleted symbols

measurements/app.py
```diff
@@ -33,15 +32,15 @@ def init_app(app):
         app.logger.setLevel(logging.WARNING)
     elif app.config['APP_ENV'] == 'development':
         app.logger.setLevel(logging.DEBUG)
-
-    # Set the jinja templates to reload when in development
-    if app.config['APP_ENV'] == 'development':
+        # Set the jinja templates to reload when in development
         app.jinja_env.auto_reload = True
         app.config['TEMPLATES_AUTO_RELOAD'] = True
+        app.config['DEBUG'] = True
```
bassosimone: Set debug when in development

measurements/app.py
```diff
@@ -33,15 +32,15 @@ def init_app(app):
...
 
     for key in app.config.keys():
-        # Do not log, even in debug, anything containing the work "SECRET"
-        if "SECRET" in key:
+        SECRET_SUBSTRINGS = ["_SECRET_", "DATABASE_URL"]
+        # Do not log, even in debug, anything containing the word "SECRET" or "DATABASE_URL"
+        if any([s in key for s in SECRET_SUBSTRINGS]):
```
bassosimone: Do not log database url.

bassosimone: Okay, verified with Python this does what it is advertised to do.

measurements/app.py
```diff
@@ -52,17 +51,21 @@ def init_app(app):
 
     cache.init_app(app, config=app.config['CACHE_CONFIG'])
 
+def check_config(config):
+    if not config['REPORTS_DIR'].startswith("s3://"):
+        raise Exception("REPORTS_DIR must start with s3://")
+
```
bassosimone: Make sure the reports dir is in `s3`

measurements/cli/__init__.py
```diff
@@ -26,6 +26,12 @@ def shell():
 
         code.interact(banner='', local=current_app.make_shell_context())
 
+@cli.command()
+@with_appcontext
+def create_tables():
+    from measurements.database import Base
+    from measurements import models
```
bassosimone: Is `models` here imported for some side effect?

measurements/database.py
```diff
@@ -15,10 +15,6 @@
 
 Base = declarative_base()
 
-def create_tables(app):
-    from measurements import models
-    Base.metadata.create_all(bind=app.db_engine)
```
bassosimone: Code moved in `cli/__init__.py`

measurements/cli/__init__.py
bassosimone: Code moved nearly verbatim from `database.py`

bassosimone: Cool: the `@cli.command` is such that `create-tables` is a command.

measurements/config.py
```diff
@@ -25,8 +23,6 @@
 # or an s3 endpoint (ex. s3://ooni-public/sanitised)
 REPORTS_DIR = os.environ.get("REPORTS_DIR", "s3://ooni-public/sanitised/")
 
-WORKERS = 4
-
```
bassosimone: Remove dead variables.

measurements/pages/__init__.py
```diff
@@ -225,36 +224,14 @@ def files_download(filename):
...
-    except S3NotConfigured:
-        raise HTTPException("S3 is not properly configured")
-
-    response = Response(
-        stream_with_context(file_chunks['content']),
-        mimetype='text/json'
     )
-    response.headers.add('Content-Length',
-                         str(file_chunks['content_length']))
-    return response
```
bassosimone: All this should be a side effect of the fact that now only s3 is supported.

measurements/models.py
```diff
@@ -121,43 +121,3 @@ class Software(Base):
     software_version = Column(String)
 
     reports = relationship("Report", back_populates="software")
-
-class ReportFile(Base):
```
bassosimone: Note: according to `git grep` this class was unused in `master`.

.travis.yml
```diff
@@ -37,9 +37,12 @@ script:
 # Wait for postgres database to be running
 - 'while ! echo exit | nc localhost 5433; do sleep 10; done'
 - sleep 10
+- make APP_ENV=testing create-tables
```
bassosimone: Okay, this creates the tables when we're running tests.

.travis.yml
```diff
@@ -37,9 +37,12 @@ script:
...
-  script: 'scripts/deploy.sh staging $HOME/.ssh/id_rsa_travis'
-  on:
-    branch: master
+
+# Disable auto deployment
+#deploy:
+#  provider: script
+#  script: 'scripts/deploy.sh staging $HOME/.ssh/id_rsa_travis'
+#  on:
+#    branch: master
```
bassosimone: Unclear to me why this was commented out, will check specific commits for hints.

Makefile
```diff
@@ -36,9 +36,12 @@ serve-d: .state/docker-build-$(APP_ENV)
 serve: .state/docker-build-$(APP_ENV)
 	$(DOCKER_COMPOSE) up
 
-debug: .state/docker-build-$(APP_ENV)
+shell: .state/docker-build-$(APP_ENV)
```
bassosimone: Ah, by reading the whole file, now I understand why this is called `shell`.

Makefile
```diff
@@ -36,9 +36,12 @@ serve-d: .state/docker-build-$(APP_ENV)
...
 	$(DOCKER_COMPOSE) run --service-ports web python -m measurements shell
 
+create-tables: .state/docker-build-$(APP_ENV)
+	$(DOCKER_COMPOSE) run web python -m measurements create_tables
+
```
bassosimone: Ah, I understand: now to create tables one needs to use the specific command. This has probably been done to avoid the warning that every worker was trying to create tables.

measurements/app.py
```diff
@@ -52,17 +51,21 @@ def init_app(app):
...
 def create_app(*args, **kw):
     from measurements import views
 
     app = Flask(__name__)
 
     # Order matters
     init_app(app)
+    check_config(app.config)
+
     init_db(app)
```
bassosimone: Predictably, `init_db` only establishes a "connection".

.travis.yml
hellais: This was used for doing continuous delivery via travis to staging.measurements.ooni.io. Since the measurement API now requires a populated database to re-enable this we either need to have a staging snapshot of it or come up with some better way to do continous delivery to staging (if we even care to have it).

For the moment I left it commented out as a reminder that it's there and it's a thing that we may care to do in the future.

Makefile
hellais: correct

measurements/cli/__init__.py
hellais: yeah the goal of importing models in here is so that the classes are created (which have sqlalchemy metaclassing magic to register something inside of the base class) so that we can then run `create_all`

measurements/models.py
hellais: Correct. It was left lying around hence I zapped it.

measurements/pages/__init__.py
hellais: Correct.

darkk:
@hellais is it possible to verify that the model still reflects DB tables during startup? I assume that DB initialization code that creates missing tables may have alike no-op codepath but I'm not familiar with this ORM.

hellais:
> @hellais is it possible to verify that the model still reflects DB tables during startup? I assume that DB initialization code that creates missing tables may have alike no-op codepath but I'm not familiar with this ORM.

Yeah it's possible to do that. Here are some hints as to you could implement that in sqlachemy: https://stackoverflow.com/questions/30428639/check-database-schema-matches-sqlalchemy-models-on-application-startup.

darkk:
@hellais it may be nice addition for applications running on top of shared DB. OTOH, version number may be good enough. Anyway, I suggest to postpone decision regarding this power-on-self-test till the first incident unless you have better ideas :)

PR #28, opened 2017-08-22
Feature/download url

hellais:
This depends on https://github.com/TheTorProject/ooni-measurements/pull/27

It adds support for using the new pipeline for serving download URLs


Dockerfile
```diff
@@ -15,25 +15,25 @@ RUN set -x \
...
 
 # Install NPM dependencies
 RUN set -x \
     && npm install -g gulp-cli
 
-# This depedency was creating issues so we install it separately
-RUN set -x \
-    && cd /tmp \
-    && npm install node-sass
-
```
bassosimone: We don't need this dependency anymore?

config/development.env
```diff
@@ -1,3 +1,3 @@
 APP_ENV=development
-REPORTS_DIR=/data/reports/
+REPORTS_DIR=s3://ooni-public/sanitised
```
bassosimone: My understanding is that this PR should stop using `s3` in favor of the pipeline. Shouldn't this mean that we should stop using `s3://` URLs as part of that?

config/testing.env
```diff
@@ -1,3 +1,3 @@
 APP_ENV=testing
-REPORTS_DIR=/data/reports/
+REPORTS_DIR=s3://ooni-public/sanitised
```
bassosimone: Ditto

measurements/api/measurements.py
```diff
@@ -140,10 +139,12 @@ def api_list_report_files():
     }
     results = []
     for row in q:
-        bucket_date, filename = row.textname.split('/')
-        url = get_download_url(current_app, bucket_date, filename)
+        download_url = urljoin(
+            current_app.config['BASE_URL'],
+            '/files/download/%s' % row.textname
```
bassosimone: What is `textname`? The textual name / hash / id of a report?

measurements/filestore.py
```diff
@@ -1,12 +0,0 @@
...
-
-from six.moves.urllib.parse import urlparse, urljoin
-
-from sqlalchemy import exists
-
-def get_download_url(app, bucket_date, filename):
-    url = "https://s3.amazonaws.com/"
-    # strip the leading s3://
-    url += app.config['REPORTS_DIR'][5:]
-    return urljoin(url, bucket_date, filename)
```
bassosimone: Okay, this is not necessary anymore. I am going to `git grep REPORTS_DIR` to understand whether it makes sense to keep it, or not.

measurements/templates/files/list.html
```diff
@@ -126,7 +126,7 @@
         {{ report_file.test_name }}
       </td>
       <td>
-        <a href="/files/download/{{ report_file.textname.split('/')[-1] }}">Download</a>
```
bassosimone: I do not have enough context to understand why here we don't split anymore.

measurements/pages/__init__.py
```diff
@@ -204,34 +208,69 @@ def files_in_country(country_code):
...
+            Measurement.intra_off.label('intra_off'),
+            Measurement.intra_size.label('intra_size'),
+            Report.report_no.label('r_report_no'),
+            Report.autoclaved_no.label('r_autoclaved_no'),
+            Autoclaved.filename.label('a_filename'),
+            Autoclaved.autoclaved_no.label('a_autoclaved_no'),
+    ).filter(Report.textname == textname) \
+        .join(Report, Report.report_no == Measurement.report_no) \
+        .join(Autoclaved, Autoclaved.autoclaved_no == Report.autoclaved_no) \
+        .order_by(Measurement.frame_off.asc(), Measurement.intra_off.asc())
```
bassosimone:
```SQL
SELECT Measurement.id, Measurement.report_no, frame_off, frame_size, intra_off, intra_size, Report.report_no, Report.autoclaved_no, Autoclaved.filename, Autoclaved.autoclaved_no FROM Measurement, Report, Autoclaved WHERE Report.report_no == Measurement.report_no AND Autoclaved.autoclaved_no == Report.autoclaved_no ORDER BY Measurement.frame_off, Measurement.intra_ff ASC```?

File: measurements/pages/__init__.py [hunk 8]
```diff
@@ -204,34 +208,69 @@ def files_in_country(country_code):
...
-    )
+    autoclaved_filename = msmts[0].a_filename
+    intra_off = msmts[0].intra_off
+    intra_size = 0
+    for msmt in msmts:
+        if autoclaved_filename != msmt.a_filename:
+            raise HTTPException("Autoclaved file is spanned across multiple files")
+        intra_size += msmt.intra_size
+    frames_off = msmts[0].frame_off
+    frames_size = msmts[-1].frame_off - frames_off + msmts[-1].frame_size
```
bassosimone: This means that measurements:

- do not span autoclaved files
- start at `frames_off`
- I do not fully understand `frame_size` (is the length starting from off, like in `struct iovec`)?
- inside a frame there is an offset `intra_off`
- I don't get `intra_size` (shouldn't it be the size of a frame minus the offset?)

measurements/pages/__init__.py
```diff
@@ -204,34 +208,69 @@ def files_in_country(country_code):
...
+        url = urljoin(current_app.config['AUTOCLAVED_BASE_URL'], autoclaved_filename)
+
+        current_app.logger.debug("Fetching autoclaved from: %s" % url)
+        current_app.logger.debug("  frames_off: %d" % frames_off)
+        current_app.logger.debug("  frames_size: %d" % frames_size)
+        current_app.logger.debug("  intra_off: %d" % intra_off)
+        current_app.logger.debug("  intra_size: %d" % intra_size)
+        try:
+            r = requests.get(url, headers={"Range": "bytes={}-{}".format(frames_off, frames_off + frames_size)},
+                                stream=True)
```
bassosimone:
```
+----------------------------------------+
| compressed file using lz4              |
+----------------------------------------+
    ^                            ^
  offset               offset + size
```

measurements/api/measurements.py
darkk: `textname` is textual name of a report file.
`filename` refers to blob storing that file: *.tar.lz4 or *.json.lz4

measurements/pages/__init__.py
```diff
@@ -204,34 +208,69 @@ def files_in_country(country_code):
...
+            streamed_data = 0
+            for chunk in lz4framed.Decompressor(r.raw):
+                current_app.logger.debug("chunking away 1")
```
bassosimone: mmm, don't understand the /meaning/ of this log message

measurements/pages/__init__.py
```diff
@@ -204,34 +208,69 @@ def files_in_country(country_code):
...
+                d = chunk
```
bassosimone: `for d in lz4framed.Decompressor` ?

measurements/pages/__init__.py
```diff
@@ -204,34 +208,69 @@ def files_in_country(country_code):
...
+                current_app.logger.debug("chunking away 2")
+                if streamed_data == 0 and intra_off > 0:
+                    d = chunk[intra_off:]
```
bassosimone: Okay, this is offset within the initial frame. (Happy to understand!)

measurements/pages/__init__.py
```diff
@@ -204,34 +208,69 @@ def files_in_country(country_code):
...
+                if streamed_data > (intra_size + len(d)):
+                    d = chunk[:(streamed_data - intra_size)]
```
bassosimone: Not very clear check

measurements/pages/__init__.py
bassosimone: Ah, maybe it's like this:

`frame_off < intra_off < frame_off + intra_size < frame_off + frame_size`?

Dockerfile
hellais: That was actually kind of a hack to avoid npm failing to install it properly. Since I switched to using yarn the dependency is installed properly.

measurements/pages/__init__.py
bassosimone: I more or less understand, but isn't it possible to express the final clamping in a more intuitive way?

measurements/pages/__init__.py
hellais: oops. These are some debug lines I left over. Will delete.

measurements/pages/__init__.py
darkk: Have you tested it? It looks a bit fishy, but I'm unsure if I understand the logic correctly.

measurements/pages/__init__.py
bassosimone: I guess this also assumes that frames returned by the database are consecutive. Perhaps this should be documented in a comment, to help the reader understand?

(I guess I am having troubles because I'm not very familiar with the way the new pipeline stores data.)

measurements/pages/__init__.py
hellais: Kind of, except you are missing the two JOINs on the Report and Autoclaved tables.

measurements/pages/__init__.py
hellais: This is to handle the case in which we have streamed more data than what we needed (for example because the offset inside of the decompressed blob ends before the end of all the decompressed data).

Example say I wanted the decompressed bytes that go from 2 - 15

````
[ 0 1 2 3 4 5 6 7 8 9 10 ] <- compressed blob
````

becomes uncompressed:

```
[ 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20] <- uncompressed blob
```

I need to take the chunk that goes from 2-15

File: Dockerfile (hunk 1 above)
Comment (bassosimone) [reply Dockerfile]: Ah, okay, it's a dependency installed by other dependencies.

File: measurements/api/measurements.py (hunk 4 above)
Comment (bassosimone) [reply measurements/api/measurements.py]: Ah, I see: it's the name of the file inside the blog. Thanks.

File: measurements/pages/__init__.py (hunk 7 above)
Comment (bassosimone) [reply measurements/pages/__init__.py]: Okay, good!

File: measurements/pages/__init__.py (hunk 8 above)
Comment (hellais) [reply measurements/pages/__init__.py]: `frames_size` and `intra_size` are just utility variables I create to make the job easier of figuring out what I need.

`frame_size`: (notice how it has an `s` at the end versus `frame_off` having no `s`) is the total size of all the frames. If the report file I need to fetch spans multiple frames I will consider `frames_size` the total of all consecutive frames (we have the guarantee from the data format that measurements in the same report end up in consecutive frames)

`intra_size` is similar to `frames_size` except it's how much uncompressed data I need to fill up the report with goodness.

File: measurements/templates/files/list.html (hunk 6 above)
Comment (hellais) [reply measurements/templates/files/list.html]: This is actually a bug that was fixed. Since `textname` already includes the `bucket` inside of it's path, we don't need to just take the last part, but we need to specify the full path inclusive of the bucket.

File: measurements/filestore.py (hunk 5 above)
Comment (bassosimone) [reply measurements/filestore.py]: @hellais:

```
$ git branch
* feature/download_url
  master
$ git grep -n REPORTS_DIR
Readme.md:88:`REPORTS_DIRECTORY` in the configuration file.
config/development.env:2:REPORTS_DIR=s3://ooni-public/sanitised
config/example-secrets.env:11:REPORTS_DIR=s3://ooni-public/sanitised
config/production.env:2:REPORTS_DIR=s3://ooni-public/sanitised
config/staging.env:2:REPORTS_DIR=s3://ooni-public/sanitised
config/testing.env:2:REPORTS_DIR=s3://ooni-public/sanitised
measurements/app.py:55:    if not config['REPORTS_DIR'].startswith("s3://"):
measurements/app.py:56:        raise Exception("REPORTS_DIR must start with s3://")
measurements/config.py:24:REPORTS_DIR = os.environ.get("REPORTS_DIR", "s3://ooni-public/sanitised/")
```

It seems to me we can get rid of this variable, no?

Comment (hellais) [reply measurements/filestore.py]: Yeah that would be a good idea.

File: measurements/pages/__init__.py (hunk 13 above)
Comment (bassosimone) [reply measurements/pages/__init__.py]: It would help, if not to have integration tests, at least to factor out this logic into unit test cases and see whether it works in such cases or not. (It may also help a comment showing the general case and the pointers.)

Comment (bassosimone) [reply measurements/pages/__init__.py]: Mumble:

```diff
+                if streamed_data > (intra_size + len(d)):
+                    d = chunk[:(streamed_data - intra_size)]
```

Re-reading this I am still in doubt.

What you want here, is to guarantee that `streamed_data + len(d) <= intra_size`, isn't it?

bassosimone: I initially also find it confusing that you use both `chunk` and `d`. But actually I see how you need both of them because there is at least one case where you need to cut both at the beginning and at the end. 👍

measurements/pages/__init__.py
bassosimone: Mumble: so, it's correct to say that `frames` variables are compressed and `intra` variables are uncompressed? If so, would you mind renaming them? I am a bit lost with the current naming.

darkk: measurements & reports, do not span autoclaved files
LZ4 frame containing measurement starts at frames_off
Every LZ4 frame is something like ~256 Kbytes
measurements always fits single LZ4 frame
single LZ4 frame may have several measurements and other stuff (e.g. tar header)
reports may be split over several LZ4 frames

Dockerfile
```diff
@@ -15,25 +15,25 @@ RUN set -x \
     && apt-get update \
     && apt-get install curl -y \
     && curl -sL https://deb.nodesource.com/setup_6.x | bash - \
-    && apt-get install git postgresql-client bzip2 gcc g++ make libpq-dev libffi-dev nodejs --no-install-recommends -y \
+    && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+    && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
```
darkk: BTW, seems, `stable` refers to `yarn`, not to `debian`, they don't provide version-specific builds and the only alternative [dist](https://github.com/yarnpkg/releases/tree/gh-pages/debian/dists) is `rc`.

measurements/templates/files/list.html
bassosimone: okay!

measurements/pages/__init__.py
bassosimone: Thanks, @darkk, I understand. So, basically, we have two base + offset couples. One for compressed and one for uncompressed data. Correct?

bassosimone: (I guess `base + offset` is my mental bias, because it reminds me of `iovec`)

Dockerfile
bassosimone: Would installing it from sources be more predictable (and not so painful)?

bassosimone: I mean, so we can pin onto a specific tag know to work in our environment.

measurements/pages/__init__.py
darkk: > `frames_size` ... `frame_size`

Just a generic request. Please, avoid variable names having different meaning and Levenshtein distance of ~1. I'm not that smart and careful to see every single letter :)

measurements/pages/__init__.py
darkk: I think that taking some files from the pipeline (or downloading them on demand) is a reasonable way to go.
IMHO, there are several corner-cases to test:
- report fitting into single frame
- report spanning 2 frames
- report spanning >2 frames (having at least one full-sized frame)
- report aligned with frame beginning
- report aligned with frame end

measurements/pages/__init__.py
darkk: > So, basically, we have two base + offset couples. One for compressed and one for uncompressed data. Correct?

Yep. These offsets are per-measurement to be able to fetch individual measurement from compressed file and report is always stored as consequent stream of bytes, so only the first and the last quadruple matter.

measurements/pages/__init__.py
```diff
@@ -204,34 +208,69 @@ def files_in_country(country_code):
...
+            Report.report_no.label('r_report_no'),
+            Report.autoclaved_no.label('r_autoclaved_no'),
+            Autoclaved.filename.label('a_filename'),
+            Autoclaved.autoclaved_no.label('a_autoclaved_no'),
+    ).filter(Report.textname == textname) \
+        .join(Report, Report.report_no == Measurement.report_no) \
+        .join(Autoclaved, Autoclaved.autoclaved_no == Report.autoclaved_no) \
+        .order_by(Measurement.frame_off.asc(), Measurement.intra_off.asc())
+
+    msmts = q.all()
```
darkk: Is there a reason to transfer all the rows form DB to application? Is it size accounting?

measurements/pages/__init__.py
hellais: @darkk do you have an easy way of generating reports that fit these criteria? That would make it easier to write some test cases for this.

measurements/pages/__init__.py
hellais: I couldn't come up with a better way of obtaining the first and last row without doing 2 queries. It seems slightly more efficient to waste some bandwidth rather than wasting CPU cycles doing two queries.

measurements/pages/__init__.py
hellais: > Mumble: so, it's correct to say that frames variables are compressed and intra variables are uncompressed? If so, would you mind renaming them? I am a bit lost with the current naming.

I used this naming as it's consistent with the naming inside of the DB schema. I think naming them differently would not lead to further clarity as it's something fairly specific (i.e. you need to know the schema of the DB to know what this is doing).

I can renames `frames_size` to `all_frames_size` (and `intras_size` to `all_intras_size`) if that makes it clearer though.

Dockerfile
hellais: I think installing yarn "from sources", which actually means installing it with `npm` is quite suboptimal for various reasons:
* Insecure fetching of resources
* Broken handling of dependencies

Which are actually the reasons why we use `yarn` instead of `npm` in the first place.

Given the fact that the whole thing is shipped as a docker container, here what we are trying to do is just reduce the likelyhood of breakages not actually have a full bullet proof solution.

Also if you are doing this building often, since these are the beginning of the dockerfile without changing instructions before that you will most likely end up with this diff of the image cached.

measurements/pages/__init__.py
darkk: The question is not wasting CPU **OR** bandwidth. I think, that the most time-consuming part of "bandwidth" consuming process is deserialization of data to python structures. I'm not motivated enough to test that, but here is some query if you want to :-)

```sql
SELECT * FROM (
  SELECT
    filename, frame_off, frame_size, intra_off, intra_size,
    row_number() OVER (ORDER BY frame_off, intra_off) AS rn,
    count(*) OVER () AS total_count,
    sum(intra_size + 1) OVER () AS report_size
  FROM autoclaved JOIN report USING (autoclaved_no) JOIN measurement USING (report_no)
  WHERE textname = '2017-05-17/20170516T205044Z-FR-AS34572-web_connectivity-20170516T205043Z_AS34572_HINf1L4OfIdyHkaDQ6WFiYaPkuIpHYeWRkigik4X66YueNOtGt-0.2.0-probe.json'
) t1 WHERE rn = 1 OR rn = total_count;
```

The tricky part is cutting the tail precisely. As far as I see, there are two

hellais: Well that would be CPU "wasted" on the side of the measurements API not on the database, which is what I am actually most concerned by.

In any case this performs adequately and works, so I don't have plans on changing it at the moment until there is evidence that another working approach yields a significant performance increase.

measurements/pages/__init__.py
darkk: I'd suggest to seek for existing blobs in the database, tell me if you need help with that. IMHO that would make test way more end-to-end.

measurements/pages/__init__.py
darkk: > significant performance increase

There is bad case having ~10k measurements within single report file, `2014-11-22/20141122T040940Z-US-AS1968-tcp_connect-no_report_id-0.1.0-probe.yaml`. Please, check if it has acceptable performance.

measurements/pages/__init__.py
bassosimone: > I used this naming as it's consistent with the naming inside of the DB schema.

Okay, this is a good argument.

> I can renames frames_size to all_frames_size (and intras_size to all_intras_size) if that makes it clearer though.

Keeping same names as database is probably wiser.

I think that a comment just saying that there is a `offset` and a `size` for compressed data and similarly a `offset` and a `size` for uncompressed data will be enough to clarify.

measurements/pages/__init__.py
bassosimone: @darkk wrote:
> IMHO, there are several corner-cases to test:
>
> - report fitting into single frame
> - report spanning 2 frames
> - report spanning >2 frames (having at least one full-sized frame)
> - report aligned with frame beginning
> - report aligned with frame end

Yes, I cannot think of more corner cases to test.

Dockerfile
bassosimone: Alright. In any case, it would not probably break as long as we stay fairly current.

measurements/pages/__init__.py
bassosimone: @darkk:

> The tricky part is cutting the tail precisely. I assume that all report files have measurements delimited by single \n (that's why intra_size + 1)

What if we silently break this assumption by mistake? That is, say on Windows I make a programming mistake in MK such that I write `\n` that the standard library converts to `\r\n` because I forgot about binary format?

> There is bad case having ~1'000'000 measurements within single report file

Fair point. I wonder how many files are similar to that. Do you have a statistic? If that is fairly common, we probably have a problem. (Is it possible to tweak the pipeline to avoid producing these corner cases, or it is intrinsic and my question does not make sense because I don't understand the pipeline?)

darkk: > What if we silently break this assumption by mistake.

The very best way is to store some checksum in database and verify it on the fly. But there is no alike checksum right now, I've not stored it as this checksum is not "immutable" and may vary if we re-sanitise private files. I think that acceptable way is to verify that served file starts with `{` and ends with `}\n`. IMHO, that's sufficient basic protection.

> I wonder how many files are similar to that.

That's the biggest one, but it should probably be still downloadable and should not lead to DoS. The second prize is for report with ~70'000 measurements.

hellais: Ok, I ended up implementing your method in the latest commit. I tested it on a couple of sample reports, but not that thoroughly.

In my limited testing I didn't actually notice a considerable performance boost, but I also didn't run it against the 1M measurement entries.

I actually didn't know there were such reports (I my limited searching I only found some 10k+ reports), so I do agree that in those case it makes sense to not waste all that amount of bandwidth and CPU on the ooni-measurements side.

measurements/pages/__init__.py
```diff
@@ -204,34 +208,84 @@ def files_in_country(country_code):
...
+                for chunk in lz4framed.Decompressor(r.raw):
+                    # I am at the beginning of the stream and I need to send some
+                    # offset into the data
+                    if streamed_data == 0 and intra_off > 0:
+                        chunk = chunk[intra_off:]
+                    # Avoid streaming more data than I should
+                    if (streamed_data + len(chunk)) > report_size:
+                        chunk = chunk[:abs(streamed_data - report_size)]
+                    yield chunk
+                    streamed_data += len(chunk)
```
darkk: Do you want to add sanity-check here that the very first byte of the streamed data is `{` and the last two are `}\n` ? I think it may be useful to catch some sort of errors early.

measurements/pages/__init__.py
```diff
@@ -204,34 +208,84 @@ def files_in_country(country_code):
...
-            current_app,
-            bucket_date,
-            filename
-        )
-    )
+    autoclaved_filename = msmts[0].filename
+    intra_off = msmts[0].intra_off
+    frame_off = msmts[0].frame_off
+    total_frame_size = msmts[-1].frame_off - msmts[0].frame_off + msmts[-1].frame_size
+    report_size = msmts[0].report_size + 1 # We add one for the final newline
```
darkk: `sum(intra_size + 1)` != `sum(intra_size) + 1`

measurements/pages/__init__.py
darkk: > found some 10k+ reports

I just run `select report_no, count(*) as cntmsm from measurement group by report_no order by cntmsm desc limit 5;` against hkgmetadb.

tests/unit/test_autoclaved.py
```diff
@@ -0,0 +1,62 @@
...
+        'autoclaved_filename': '2016-07-07/http_invalid_request_line.0.tar.lz4',
+        'textname': '2016-07-07/20160706T015518Z-CH-AS200938-http_invalid_request_line-20160706T015558Z_AS200938_iVsUREhX4hTEoHTOfyTFXyLvXKGGd80sFE3Xw3pJIUg2TDXr9I-0.2.0-probe.json',
+        'frame_off': 0,
+        'total_frame_size': 1897,
+        'intra_off': 1536,
+        'report_size': 3118,
+        # Note this doesn't match exactly what is in sanitised
+        'expected_shasum': 'e13999959d636ad2b5fd8d50493a257a2c616b0adad086bf7211de5f09463f6d'
+    }
+}
```
darkk: Do you need help mining other suggested cases?

tests/unit/test_autoclaved.py
```diff
@@ -0,0 +1,62 @@
+import hashlib
+from measurements.pages import decompress_autoclaved
+
+# The s3 URL is something like:
+# https://s3.amazonaws.com/ooni-public/sanitised/2017-07-08/20170707T051609Z-IR-AS44244-web_connectivity-20170707T051545Z_AS44244_ujwFYcBJXcL2MjZnHXhBqVEOG2iHR0AuPUHB7aGGqUhdtvz70h-0.2.0-probe.json
+
+AUTOCLAVED = {
+    # This is the case of a report spanning many frames
+    # XXX this report is actually quite big maybe we want to pick as a test
```
darkk: > quite big

I think that downloding blobs on demand is probably better than placing them into the repo.

measurements/pages/__init__.py
hellais: Ah right, because the `intra_size` is not inclusive of the `\n` character. Will adjust accordingly.

measurements/pages/__init__.py
hellais: I am going to add a sanity check for starting with `{` and ending with `\n`. I will not match against the two last bytes to avoid having to handle buffering (maybe the last chunk the decoder serves me contains only 1 byte).

measurements/pages/__init__.py
```diff
@@ -204,34 +208,93 @@ def files_in_country(country_code):
...
+                    if (streamed_data + len(chunk)) > report_size:
+                        chunk = chunk[:abs(streamed_data - report_size)]
+                    yield chunk
+
+                    # Sanity checks to ensure the streamed data start with
+                    # `{` and ends with `\n`
+                    if streamed_data == 0:
+                        assert chunk[:1] == b'{', 'Chunk starts with %s != {' % chunk[:1]
+                    if streamed_data + len(chunk) == report_size:
+                        assert chunk[-1:] == b'\n', 'Chunk ends with %s != \\n' % chunk[-1:]
```
darkk: IMHO `%s` should be replaced with `%r` (repr) as bad bytes may be binary.

tests/unit/test_autoclaved.py
hellais: If you have a quick and easy way to do it that would be good.

measurements/pages/__init__.py
```diff
@@ -204,34 +208,93 @@ def files_in_country(country_code):
...
+    q = current_app.db_session.query(
+        subquery.c.frame_off,
+        subquery.c.frame_size,
+        subquery.c.intra_off,
+        subquery.c.intra_size,
+        subquery.c.row_number.label('c_row_number'),
+        subquery.c.total_count.label('c_total_count'),
+        subquery.c.report_size,
+        subquery.c.filename
+    ).filter(or_(subquery.c.row_number == 1, subquery.c.row_number == subquery.c.total_count))
```
bassosimone: What is this exactly?

bassosimone: A way to avoid fetching all the rows?

bassosimone: Ah, yes, now I see that it matches what @darkk suggested

measurements/pages/__init__.py
```diff
@@ -215,32 +215,31 @@ def decompress_autoclaved(
...
-                    # offset into the data
-                    if streamed_data == 0 and intra_off > 0:
-                        chunk = chunk[intra_off:]
-                    # Avoid streaming more data than I should
-                    if (streamed_data + len(chunk)) > report_size:
-                        chunk = chunk[:abs(streamed_data - report_size)]
-                    yield chunk
+            beginning = True
+            # Create a copy because we are in a closure
+            to_read = report_size
```
darkk: I thought that floats() are immutable... I don't understand the reason to make a copy :-(

measurements/pages/__init__.py
```diff
@@ -215,32 +215,31 @@ def decompress_autoclaved(
...
-                    streamed_data += len(chunk)
-            assert report_size == streamed_data, 'I expected to stream %d but only did %d' % (report_size, streamed_data)
+                    if beginning:
+                        assert d[:1] == b'{', 'Chunk starts with %r != {' % d[:1]
+                    if to_read == len(d):
+                        assert d[-1:] == b'\n', 'Chunk ends with %r != \\n' % d[-1:]
+
+                    yield d
+                    to_read -= len(d)
+                    beginning = False
```
darkk: How does caller handle zero-sized chunks? Does caller treat them as EOF or not?

measurements/pages/__init__.py
hellais: That's how closures work in python. As soon as you assign the enclosed variable (as we do below in `to_read -= len(d)`) it's converted to a local reference and since that conversion is happening within the scope of the closure the `while` loop will given an unboundvariable error.

measurements/pages/__init__.py
hellais: EOF is determined by the generator exiting. I believe there is a guarantee that no zero sized chunk will be spit out by the `lz4framed.Decompressor`: https://github.com/Iotic-Labs/py-lz4framed/blob/master/lz4framed/py-lz4framed.c#L880

measurements/pages/__init__.py
```diff
@@ -204,34 +208,92 @@ def files_in_country(country_code):
...
+    q = current_app.db_session.query(
+        subquery.c.frame_off,
+        subquery.c.frame_size,
+        subquery.c.intra_off,
+        subquery.c.intra_size,
+        subquery.c.row_number.label('c_row_number'),
+        subquery.c.total_count.label('c_total_count'),
+        subquery.c.report_size,
+        subquery.c.filename
+    ).filter(or_(subquery.c.row_number == 1, subquery.c.row_number == subquery.c.total_count))
```
bassosimone: @hellais why `subquery.c.row_number` and not `subquery.c_row_number`?

tests/unit/test_unit.py
```diff
@@ -30,18 +30,6 @@ def test_api_docs(client):
...
-    assert resp.status_code == 200
-
-def test_api_private_counties_by_month(client):
-    resp = client.get(url_for('api_private.api_private_counties_by_month'))
-    assert resp.status_code == 200
-
-def test_api_private_runs_by_month(client):
-    resp = client.get(url_for('api_private.api_private_runs_by_month'))
-    assert resp.status_code == 200
-
```
bassosimone: Why have these tests been deleted?

tests/unit/test_autoclaved.py
```diff
@@ -0,0 +1,62 @@
...
+    # something smaller that still proves the point or maybe we write some code
+    # to generate it
+    'many_frames': {
+        'autoclaved_filename': '2017-07-08/20170707T051609Z-IR-AS44244-web_connectivity-20170707T051545Z_AS44244_ujwFYcBJXcL2MjZnHXhBqVEOG2iHR0AuPUHB7aGGqUhdtvz70h-0.2.0-probe.json.lz4',
+        'textname': '2017-07-08/20170707T051609Z-IR-AS44244-web_connectivity-20170707T051545Z_AS44244_ujwFYcBJXcL2MjZnHXhBqVEOG2iHR0AuPUHB7aGGqUhdtvz70h-0.2.0-probe.json',
+        'frame_off': 0,
+        'total_frame_size': 15485694 + 1724,
+        'intra_off': 0,
+        'report_size': 74695236,
+        # This doesn't actually match that which is in s3
```
bassosimone: I do not understand the purpose of this comment? Also, below it's `autoclaved`, here `s3`, is that intentional?

tests/unit/test_unit.py
hellais: Because they were failing non-deterministically (due to the use of the caching flask endpoint, which we should also BTW stop using) and are in any case unittests for private API methods that I don't think we should be unittesting anyways.

bassosimone: Fair enough.

darkk:
> :whale:

[kek](https://en.wikipedia.org/wiki/Blue_Whale_(game))

PR #30, opened 2017-09-04
Openapi

hellais:
Screenshot of how the new docs look like:

screen shot 2017-09-05 at 16 51 27


measurements/api/__init__.py
```diff
@@ -1,3 +1,5 @@
-from .docs import api_docs_blueprint
 from .private import api_private_blueprint
-from .measurements import api_blueprint
+
+from .measurements import get_version
+from .measurements import list_files
+from .measurements import list_measurements, get_measurement
```
bassosimone: Why here you're just changing the imports?

bassosimone:
@hellais so, this is basically changing the code to use a `yml` based API specification that allows, among other things, for more easy arguments binding and for automatic documentation generation?

hellais:
> @hellais so, this is basically changing the code to use a yml based API specification that allows, among other things, for more easy arguments binding and for automatic documentation generation?

Correct.

Gulpfile.babel.js
```diff
@@ -24,6 +24,10 @@ let modulesPath = path.join(basePrefix, "node_modules");
 let webpackConfig = {
   quiet: false,
   verbose: true,
+  node: {
+    fs: "empty",
+    child_process: "empty"
+  },
```
bassosimone: I assume this change was automatically generated, correct?

measurements/api/__init__.py
bassosimone: My understanding is that `api_blueprint` was a global object containing the routes. Since now we use another approach in which routing is implemented using a `YAML` file, we get directly the names of the APIs rather than getting them through the blueprint wrapper.

Also, these names are imported here &mdash; and not used &mdash; because here you're following the convention of making symbols available through `__init__.py` as they were part of `../api.py` but the implementation is in other files.

measurements/api/measurements.py
```diff
@@ -106,13 +86,6 @@ def api_list_report_files():
     if since_index:
         q = q.filter(Report.report_no > report_no)
 
-    # XXX these are duplicated above, refactor into function
-    if order.lower() not in ('asc', 'desc'):
-        raise BadRequest("Invalid order")
-    if order_by not in ('test_start_time', 'probe_cc', 'report_id',
-                        'test_name', 'probe_asn', 'report_no'):
-        raise BadRequest("Invalid order_by")
-
```
bassosimone: These checks have vanished. I am going to assume they're now embedded in the `YAML` file.

- [ ] make sure that's really the case.

measurements/api/measurements.py
```diff
@@ -149,15 +122,14 @@ def api_list_report_files():
             'probe_asn': "AS{}".format(row.probe_asn),
             'test_name': row.test_name,
             'index': int(row.report_no) + REPORT_INDEX_OFFSET,
-            'test_start_time': row.test_start_time.strftime('%Y-%m-%dT%H:%M:%SZ')
+            'test_start_time': row.test_start_time
```
bassosimone: Why?

bassosimone: Ah, because now you've implemented a flask filter to automatically encode into JSON.

measurements/app.py
```diff
@@ -58,6 +78,7 @@ def create_app(*args, **kw):
     from measurements import views
 
     app = Flask(__name__)
+    app.json_encoder = FlaskJSONEncoder
```
bassosimone: Does this mean that the thing you return from handlers is `JSON` encoded by flask using this filter?

(I assume that was already the case, but you had the issue that some objects were not trivially JSON encodeable, like datetime, and hence you needed to make sure they were explicitly in the code.)

measurements/openapi/measurements.yml
```diff
@@ -0,0 +1,424 @@
...
+          description: Returns version information
+          schema:
+            $ref: "#/definitions/VersionInfo"
+
+
+        default:
+          description: Default response
+
+definitions:
+  VersionInfo:
```
bassosimone: So, this is basically the thing you refer to when you do:

> $ref: "#/definitions/VersionInfo"

?

measurements/pages/docs.py
```diff
@@ -8,6 +8,4 @@
 
 @api_docs_blueprint.route('/', methods=["GET"])
 def api_docs():
-    with open(os.path.join(BASE_DIR, 'markdown', 'api_docs.md')) as in_file:
-        docs_text = in_file.read()
-    return render_template('api.html', docs_text=docs_text)
+    return render_template('api.html')
```
bassosimone: Does this mean `api_docs.md` can also be removed?

hellais: Yes `api_docs` can be remove. I have not done it yet, because I want to first ensure I haven't missed anything.

measurements/openapi/measurements.yml
hellais: Correct.

measurements/app.py
hellais: Yes, but I extend it with the serialisation of datetime to avoid duplicating code in the handlers.

measurements/api/measurements.py
hellais: yup

measurements/api/__init__.py
hellais: Correct.

Gulpfile.babel.js
hellais: No I had to add it to fix breakage in some node packages.

bassosimone: ah, okay

measurements/app.py
bassosimone: perfect

measurements/pages/docs.py
bassosimone: okay, I assume this can also be done in another PR

PR #31, opened 2017-09-05
Feature/anomalies

hellais:
This is a work in progress branch to implement anomaly features inside of measurements API.

It's currently blocking on having anomalies extracted and inside of the database.

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -263,7 +263,7 @@ def list_measurements(
 
     q = current_app.db_session.query(*cols)\
             .join(Report, Report.report_no == Measurement.report_no)\
-            .join(Label, Label.msm_no== Measurement.msm_no)
+            .outerjoin(Label, Label.msm_no== Measurement.msm_no)
 
     if input_:
         q = q.join(Measurement, Measurement.input_no == Input.input_no)
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -257,6 +257,8 @@ def list_measurements(
             'test_name': row.test_name,
             'measurement_start_time': row.measurement_start_time,
             'input': row.input if row.m_input_no else None,
+            # XXX @darkk how do I get this field populated?
+            'anomaly_type': None
         })
 
     pages = -1
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -294,7 +294,7 @@ def list_measurements(
         q = q.filter(Measurement.measurement_start_time <= until)
 
     if confirmed is not None:
-        q = q.filter(c_confirmed== confirmed)
+        q = q.filter(c_confirmed == confirmed)
     if anomaly is not None:
         q = q.filter(c_anomaly == anomaly)
     if failure is True:
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -159,6 +159,9 @@ def get_measurement(measurement_id):
     except exc.MultipleResultsFound:
         current_app.logger.warning("Duplicate rows for measurement_id: %s" % measurement_id)
         msmt = q.first()
+    except exc.NoResultFound:
+        # XXX we should actually return a 404 here
+        raise BadRequest("No measurement found")
 
     r = requests.get(urljoin(current_app.config['AUTOCLAVED_BASE_URL'], msmt.a_filename),
             headers={"Range": "bytes={}-{}".format(msmt.frame_off, msmt.frame_off+msmt.frame_size)}, stream=True)
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -12,7 +12,7 @@ from flask import Blueprint, current_app, request
 from flask.json import jsonify
 from werkzeug.exceptions import HTTPException, BadRequest
 
-from sqlalchemy import func
+from sqlalchemy import func, or_, and_
 from sqlalchemy.orm import lazyload, exc
 
 from six.moves.urllib.parse import urljoin, urlencode
@@ -163,6 +163,13 @@ def get_measurement(measurement_id):
         mimetype=current_app.config['JSONIFY_MIMETYPE']
     )
 
+def clean_list_arg(l):
+    """
+    Takes a list argument in the form of "foo, bar, tar" and returns it as a
+    cleaned up list (["foo", "bar", "tar"]).
+    """
+    return map(lambda x: x.strip(), l.split(','))
+
 def list_measurements(
         report_id=None,
         probe_asn=None,
@@ -174,7 +181,10 @@ def list_measurements(
         order_by=None,
         order='desc',
         offset=0,
-        limit=100
+        limit=100,
+        failure=None,
+        anomaly=None,
+        confirmed='true'
     ):
     input_ = request.args.get("input")
 
@@ -182,6 +192,25 @@ def list_measurements(
         if probe_asn.startswith('AS'):
             probe_asn = probe_asn[2:]
         probe_asn = int(probe_asn)
+
+    # When the user specifies a list that includes all the possible values for
+    # boolean arguments, that is logically the same of applying no filtering at
+    # all.
+    if failure is not None:
+        if set(clean_list_arg(failure)) == set(['true', 'false']):
+            failure = None
+        else:
+            failure = failure == 'true'
+    if anomaly is not None:
+        if set(clean_list_arg(anomaly)) == set(['true', 'false']):
+            anomaly = None
+        else:
+            anomaly = anomaly == 'true'
+    if confirmed is not None:
+        if set(clean_list_arg(confirmed)) == set(['true', 'false']):
+            confirmed = None
+        else:
+            confirmed = confirmed == 'true'
 
     try:
         if since is not None:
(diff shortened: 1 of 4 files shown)


hellais:
xref: https://github.com/TheTorProject/ooni-pipeline/issues/76

darkk:
Is it possible to have `confirmed=true` while `anomaly=false`?

measurements/api/measurements.py
```diff
@@ -235,6 +281,24 @@ def list_measurements(
     if until is not None:
         q = q.filter(Measurement.measurement_start_time <= until)
 
+    if confirmed is not None:
+        q = q.filter(c_confirmed== confirmed)
```
darkk: Whiespace before `==` :)

hellais:
> Is it possible to have confirmed=true while anomaly=false?

Yeah this is possible. A case of this sort will happen when the on-device anomaly detection says that it's not a network anomaly, but the pipeline found a fingerprint in the probe.

We could maybe "enforce" that when it's confirmed, then the anomaly bit is automatically flipped, but I would say it's best to do that inside of a separate table as it's actually useful to be able to quickly search for these sorts of measurements to improve the on-device heuristics.

measurements/api/measurements.py
```diff
@@ -143,26 +151,40 @@ def get_measurement(measurement_id):
...
         msmt = q.one()
     except exc.MultipleResultsFound:
         current_app.logger.warning("Duplicate rows for measurement_id: %s" % measurement_id)
         msmt = q.first()
+    except exc.NoResultFound:
+        # XXX we should actually return a 404 here
+        raise BadRequest("No measurement found")
 
     r = requests.get(urljoin(current_app.config['AUTOCLAVED_BASE_URL'], msmt.a_filename),
             headers={"Range": "bytes={}-{}".format(msmt.frame_off, msmt.frame_off+msmt.frame_size)}, stream=True)
```
darkk: Looks like another off-by-one error, it should be `frame_off + frame_size - 1` as both ends of `Range` query are included.

measurements/api/measurements.py
```diff
@@ -143,26 +151,40 @@ def get_measurement(measurement_id):
...
 
     # XXX use for streaming support lz4framed.Decompressor
     # @darkk how big can these lz4 frames even become? Should this be a concern?
     msmt_data = lz4framed.decompress(r.raw.read())[msmt.intra_off:msmt.intra_off+msmt.intra_size]
+
+    # XXX do we also want to replace on the fly the measurement_id inside of
+    # the report itself?
```
darkk: I don't.
I think that client-generated `measurement_id` is valuable for idempotency. It's just not 100% unique. It's like `Message-Id` in emails :)

measurements/api/measurements.py
```diff
@@ -183,6 +208,25 @@ def list_measurements(
...
+        if set(clean_list_arg(anomaly)) == set(['true', 'false']):
+            anomaly = None
+        else:
+            anomaly = anomaly == 'true'
+    if confirmed is not None:
+        if set(clean_list_arg(confirmed)) == set(['true', 'false']):
+            confirmed = None
+        else:
+            confirmed = confirmed == 'true'
+
```
darkk: I don't get the logic. `probe_asn` is always SINGLE value, while `failure` is boolean and may be list of values. It looks like some inconsistency to me. What's the reason of it? IMHO, there should be some comment on that if that's intentional.

measurements/api/measurements.py
```diff
@@ -244,19 +323,25 @@ def list_measurements(
...
+        measurement_id = '{}-{}'.format(MSM_ID_PREFIX, row.msm_no)
         url = urljoin(
             current_app.config['BASE_URL'],
-            '/api/v1/measurement/%s' % row.m_id
+            '/api/v1/measurement/%s' % measurement_id
         )
         results.append({
             'measurement_url': url,
-            'measurement_id': row.m_id,
+            'measurement_id': measurement_id,
```
darkk: IMHO we should not override `measurement_id` field here according to principle of least astonishment. There is already entity named `measurement_id` and it's user-generated.

darkk:
> Yeah this is possible.

Well, I deduced that the combination is impossible from API schema. I'm unsure that the idea you suggest is a "correct" behavior... Maybe. Although it's a bit unexpected.

measurements/api/measurements.py
hellais: I think you mean to say `failure` or `anomaly`. In any case the logic here is that I may want to look for `anomaly == True || anomaly == False`, which is different than asking for `anomaly == True`. That is encoded explicitly as `?anomaly=true,false`. Effectively that is equivalent to not imposing any filtering at all.

I can clarify this in the comment though.

measurements/api/measurements.py
hellais: Following the principle of least astonishment, I would expect that I can use `measurement_id` as some identifying to identify measurements, that is a thing I cannot do.

IMHO we should be transforming this and adding it on the pipeline stage. See: https://github.com/TheTorProject/ooni-pipeline/issues/48#issuecomment-328652712. For lack of that I would say we rewrite it on the measurements side for consistency purposes.

measurements/api/measurements.py
hellais: If we don't there is a mismatch between this and what we return below. I think that having `measurement_id` being user generated has basically null value to an end user of our data, since they can't really do anything with it. I would at this point either strip it from the data, or rewrite it with something slightly more meaningful.

hellais:
> Well, I deduced that the combination is impossible from API schema. I'm unsure that the idea you suggest is a "correct" behavior... Maybe. Although it's a bit unexpected.

What made you deduce this from the API schema? The API schema has these things on separate query parameters precisely to give an indication of the fact that this combination is possible.

Example:

`/measurements?confirmed=true&anomaly=false`

because the client side detection logic returned true, BUT a fingerprint was found.

We could also not make it like this, but this would mean enforcing this on the pipeline side of things, kind of making `confirmed` a "more powerful" `anomaly`.

measurements/api/measurements.py
darkk: > I would expect that I can use measurement_id as some identifying to identify measurements

Well, it's possible. You can use it to identify some set of measurements (usually -- one measurement).

> that is a thing I cannot do

I'd rather drop `measurement_id` from results then. Or just name it differently. I mean, there is entity named `measurement_id` in measurement json blob and it's there till last "old" ooniprobe dies, so we can't easily rename it.

Values having same name should have same semantics. Do you agree with that?

Let's just name it `measurement_gooid`*. I use `oo` from _OONI_ and I don't want it to be confused with 128-bit `GUID` format. OTOH, _goo-id_ is probably too self-critical. :)

What do you think?

hellais:
Yes gooid is goouid!

hellais:
I mean we rename current measurement_id to gooid and keep measurement_id for actual id. Is this what you meant?
We anyways do similar things for report_id

measurements/api/measurements.py
```diff
@@ -174,7 +196,10 @@ def list_measurements(
         order_by=None,
         order='desc',
         offset=0,
-        limit=100
+        limit=100,
+        failure=None,
+        anomaly=None,
+        confirmed=None
     ):
     input_ = request.args.get("input")
```
darkk:
```
>>> def foo(input=1):
...  pass
... 
```

measurements/api/measurements.py
```diff
@@ -143,26 +151,40 @@ def get_measurement(measurement_id):
...
         msmt = q.first()
+    except exc.NoResultFound:
+        # XXX we should actually return a 404 here
+        raise BadRequest("No measurement found")
 
     r = requests.get(urljoin(current_app.config['AUTOCLAVED_BASE_URL'], msmt.a_filename),
             headers={"Range": "bytes={}-{}".format(msmt.frame_off, msmt.frame_off+msmt.frame_size)}, stream=True)
 
     # XXX use for streaming support lz4framed.Decompressor
     # @darkk how big can these lz4 frames even become? Should this be a concern?
```
darkk: MAX is 27791334 bytes for frame, 29293666 bytes for measurements

measurements/api/measurements.py
```diff
@@ -143,26 +151,40 @@ def get_measurement(measurement_id):
...
     msmt_data = lz4framed.decompress(r.raw.read())[msmt.intra_off:msmt.intra_off+msmt.intra_size]
```
hellais: @darkk claims this can be maximum 27MB. Let's file a ticket about improving this to be done via streaming.

darkk: For the record -- there is a bug in `lz4framed.decompress`, it does not raise exception on leftover data.

measurements/api/measurements.py
darkk: Actually, I was a bit wrong. We don't have thing named `measurement_id ` in th data, we have `id` and `report_id`, so we probably can take `measurement_id` name here. It may be some sort of backwards incompatibility, but I don't feel like it matters.

measurements/api/measurements.py
darkk: Okay, here is #35

measurements/api/measurements.py
darkk: TheTorProject/ooni-pipeline#83

measurements/api/measurements.py
darkk: #36

darkk: #36

darkk:
LGTM.

PR #33, opened 2017-09-07
Add more tests for decompress_autoclaved, use smaller file for full-file test-case

darkk:
That's how I found the test-case for #28 :)

```sql
drop table if EXISTS msm_framestart;
create temporary table msm_framestart as
select autoclaved_no, msm.*
from measurement as msm
join report using (report_no)
where frame_off != 0 and intra_off = 1024; -- tar alignment + tar header for next file

select count(*) from msm_framestart;

drop table if EXISTS msm_frameend;
create temporary table msm_frameend as
select prev.*
  from msm_framestart as next
  join measurement as prev on (prev.report_no != next.report_no AND prev.frame_off + prev.frame_size = next.frame_off)
  join report as prevfile ON (prev.report_no = prevfile.report_no AND prevfile.autoclaved_no = next.autoclaved_no);

select count(*) from msm_frameend;

drop table if exists msm_framelast;
create temporary table msm_framelast as
select * from (
  select
    autoclaved_no,
    row_number() OVER (PARTITION BY autoclaved_no, frame_off ORDER BY intra_off) AS rn,
    count(*) OVER (PARTITION BY autoclaved_no, frame_off) AS total_count,
    msm.*
  from msm_frameend AS msm
  join report as prevfile USING (report_no)
) t1 where rn = total_count;

-- two frames, two different reports
select
  prev.report_no, next.report_no,
  prev.frame_off, prev.frame_size, prev.frame_off + prev.frame_size = next.frame_off,
  next.frame_off, next.frame_size,
  prev.intra_off, prev.intra_size,
  next.intra_off, next.intra_size
from msm_framestart as next
join report as nextfile ON (next.report_no = nextfile.report_no)
join msm_framelast as prev ON (nextfile.autoclaved_no = prev.autoclaved_no AND prev.frame_off + prev.frame_size = next.frame_off)
where prev.report_no != next.report_no;

create temporary table report_nl as
select
  prev.report_no
from msm_framestart as next
join report as nextfile ON (next.report_no = nextfile.report_no)
join msm_framelast as prev ON (nextfile.autoclaved_no = prev.autoclaved_no AND prev.frame_off + prev.frame_size = next.frame_off)
where prev.report_no != next.report_no;

select report_no, count(distinct frame_off) from measurement where report_no in (select * from report_nl) group by report_no;
-- all this reports are single-framed

-- Some of reports that have unusual intra_off
select textname from report where report_no in (332953, 47263); -- first one has \n\0\0\0\0 at frame beginning followed by tar header

-- Some of reports that have tail ALMOST (?) aligned with end of frame...
select report_no, textname from report where report_no in (303388, 106782);

SELECT * FROM (
  SELECT
    filename, frame_off, frame_size, intra_off, intra_size,
    row_number() OVER (ORDER BY frame_off, intra_off) AS rn,
    count(*) OVER () AS total_count,
    sum(intra_size + 1) OVER () AS report_size
  FROM autoclaved JOIN report USING (autoclaved_no) JOIN measurement USING (report_no)
  WHERE textname =
        -- '2016-04-02/20160401T114238Z-GB-AS2856-http_requests-no_report_id-0.1.0-probe.yaml' -- this one has \n\0\0\0\0 at frame beginning followed by tar header
        -- '2016-04-19/20160418T050806Z-GB-AS5607-http_requests-no_report_id-0.1.0-probe.yaml' -- this one has \n\0\0\0\0 at frame beginning followed by tar header
        '2016-04-06/20160405T142624Z-GB-AS5607-http_requests-no_report_id-0.1.0-probe.yaml'
) t1 WHERE rn = 1 OR rn = total_count;
```


package.json
```diff
@@ -1,7 +1,4 @@
 {
-  "engines": {
-    "node": "6.6.0"
-  },
```
darkk: That's probably `curl -sL https://deb.nodesource.com/setup_6.x | bash -` installing 6.11.3-1nodesource1 instead of 6.6.0 ;-)

> `Get:2 https://deb.nodesource.com/node_6.x/ jessie/main nodejs amd64 6.11.3-1nodesource1 [9881 kB]`

tests/unit/test_autoclaved.py
```diff
@@ -44,23 +52,46 @@
...
+    # of tar padding and headers, but here is something like that.
+    'next_frame_newline': {
+        'autoclaved_filename': '2016-04-18/http_requests.06.tar.lz4',
+        'textname': '2016-04-18/20160417T080443Z-GB-AS5607-http_requests-no_report_id-0.1.0-probe.yaml',
+        'frame_off': 555968,
+        'total_frame_size': 86597, # single frame, all alike reports fit into single frame :-/
+        'intra_off': 211073,
+        'report_size': 151041,
+        'expected_shasum': '58dbc014da07fb3192e2231be37e1acc7344cf87ecc8afabf84bf5c387d1a825',
+    },
```
darkk: Well, that's triggering lz4 exception. That's actually both good and bad.
It's good as it's not triggering infinite busy loop that I was afraid of.
It's bad as client will **likely** get truncated output, it's tricky to signal error in HTTP, so exception will lead to connection reset between nginx and API server and (right now) only god knows if nginx will serve EOF to client or reset connection as well and even god does not know how client's library will handle that.

measurements/pages/__init__.py
```diff
@@ -220,12 +220,13 @@ def decompress_autoclaved(
...
             url = urljoin(current_app.config['AUTOCLAVED_BASE_URL'], autoclaved_filename)
-            headers = {"Range": "bytes={}-{}".format(frame_off, frame_off + total_frame_size)}
+            # byte positions specified are inclusive -- https://tools.ietf.org/html/rfc7233#section-2.1
+            headers = {"Range": "bytes={}-{}".format(frame_off, frame_off + total_frame_size - 1)}
             r = requests.get(url, headers=headers, stream=True)
             beginning = True
             # Create a copy because we are in a closure
             to_read = report_size
-            while to_read > 0:
+            while to_read > 1:
```
hellais: This is `1` instead of `0` because `report_size` is actually include of the newline character?

measurements/pages/__init__.py
```diff
@@ -234,14 +235,22 @@ def generator():
...
 
                     yield d
                     to_read -= len(d)
-                    beginning = False
+                    if len(d): # valid lz4 frame may have 0 bytes
+                        beginning = False
+            # `autoclaved` file format may have `\n` in separate LZ4 frame,
+            # database stores offset for JSON blobs without trailing newline,
+            # here is hack adding newline as next frame boundaries are unknown.
+            if r.raw.read(1) != b'': # stream must be already EOFed
```
hellais: I thinks this should actually never happen, but I guess it's ok to check just in case (note: I was initially confused by this and thought it was somehow related to the change of the while clause being `1` instead of `0`)

measurements/pages/__init__.py
```diff
@@ -277,6 +286,7 @@ def files_download(textname):
     if len(msmts) == 0:
         current_app.logger.debug("Could not find %s" % textname)
         raise NotFound("No file with that filename found")
+    msmts.sort(key=operator.attrgetter('frame_off')) # at most two rows, but it could be single
```
hellais: Good call. I thought about the sorting, but forgot to add it.

tests/unit/test_autoclaved.py
```diff
@@ -44,23 +55,46 @@
...
-        download_size = 0
-        h = hashlib.sha256()
-        g = decompressor()
-        all_data = b''
-        for chunk in g:
-            all_data += chunk
-            h.update(chunk)
-        assert len(all_data) == ac['report_size'], len(all_data)
-        assert h.hexdigest() == ac['expected_shasum'], h.hexdigest()
+@pytest.mark.parametrize("name,ac", AUTOCLAVED.items())
```
hellais: Ah neat. I didn't know `pytest` supported this (this is actually the first time I use pytest)!

hellais:
If you answer my questions to ensure I understand all that is going on I will merge. The diff looks good.

Do you prefer I merge with a merge commit or via a rebase?

measurements/pages/__init__.py
darkk: Yep. And the last `\n` has no information regarding frame if it's separate one.

measurements/pages/__init__.py
darkk: It catched off-by-one error in Range :)

darkk:
> merge commit or rebase
I usually use merge commit to preserve history if it's more valuable than disk space :) Do we have some sort of established preference for repositories?

hellais:
> I usually use merge commit to preserve history if it's more valuable than disk space :) 

I generally go for a rebase when the history of the commits is not as important as having a concise way to document the change (I can point to a single commit that implements the change).

I guess this actually doesn't matter that much given the fact that we can reference pull requests.

> Do we have some sort of established preference for repositories?

Not, but I should probably add something about this inside of https://github.com/OpenObservatory/style

PR #39, opened 2017-09-28
Release stable measurements API

hellais:


measurements/api/measurements.py
```diff
@@ -213,20 +206,21 @@ def list_measurements(
     # boolean arguments, that is logically the same of applying no filtering at
     # all.
     if failure is not None:
-        if set(clean_list_arg(failure)) == set(['true', 'false']):
+        if set(failure) == set(['true', 'false']):
```
darkk: btw, that can be written as `{'true', 'false'}`

measurements/api/measurements.py
```diff
@@ -213,20 +206,21 @@ def list_measurements(
...
             failure = None
         else:
-            failure = failure == 'true'
+            failure = failure == ['true']
```
darkk: That's probably `set(failure) == {'true'}` because of `&failure=true&failure=true`

measurements/api/measurements.py
```diff
@@ -213,20 +206,21 @@ def list_measurements(
...
     if anomaly is not None:
-        if set(clean_list_arg(anomaly)) == set(['true', 'false']):
+        current_app.logger.warning("Duplicate rows for measurement_id: %s" % anomaly)
```
darkk: WAT? It looks like stale debug output.

measurements/api/measurements.py
```diff
@@ -244,11 +238,11 @@ def list_measurements(
...
 
-    c_anomaly = func.coalesce(Label.anomaly, Measurement.anomaly, False)\
+    c_anomaly = func.coalesce(Label.anomaly, Measurement.anomaly, false())\
                     .label('anomaly')
-    c_confirmed = func.coalesce(Label.confirmed, Measurement.confirmed, False)\
+    c_confirmed = func.coalesce(Label.confirmed, Measurement.confirmed, false())\
                     .label('confirmed')
-    c_msm_failure = func.coalesce(Label.msm_failure, Measurement.msm_failure, False)\
+    c_msm_failure = func.coalesce(Label.msm_failure, Measurement.msm_failure, false())\
                     .label('msm_failure')
```
darkk: Wow. Does it really matter if it's `False` or `false()`? It looks like obscure rake if it does.

measurements/api/measurements.py
hellais: Yes that is equivalent, however maybe it's clearer that we are comparing sets if the list is cast to a set explicitly.

measurements/api/measurements.py
hellais: oops, yes good catch!

measurements/api/measurements.py
hellais: It actually doesn't matter (I realised later). The SQL query it generates however is must more clearer (it will use `false` in the query instead of injecting it via variable substitution).
I added this while I was trying to figure out why I saw many measurements with `anomaly=true && failure=true && confirmed=true` to be sure that it was not an issue with `coalesce`, but figured I may leave it nonetheless as [sqlachemy docs claim](http://docs.sqlalchemy.org/en/latest/core/sqlelement.html#sqlalchemy.sql.expression.false) "true() and false() feature better integrated behavior within conjunctions and on dialects that don’t support true/false constants"

measurements/api/measurements.py
hellais: We don't currently support the multi variable style of passing arguments, but only the csv one (`true,false`). Currently if you pass `failure=true,true` you are right to point out that it will not actually be passed as true, but I would consider this a mis-usage of the API.

Maybe in future versions we can be more tollerant.

PR #42, opened 2017-11-20
Explorer compatibility API

darkk:


measurements/api/legacy_explorer.py
```diff
@@ -0,0 +1,135 @@
...
+
+from flask import Blueprint, current_app, request
+from flask.json import jsonify
+from sqlalchemy import func # , or_, and_, false, true
+
+from measurements.models import Report, Input, Measurement, Autoclaved
+from measurements.api.measurements import fetch_autoclaved_text
+
+# prefix: /api/legacy_explorer
+api_legacy_explorer_blueprint = Blueprint('api_legacy_explorer', 'measurements')
```
hellais: I would put these legacy endpoint inside of the private blueprint (which has the `_` prefix) and maybe call it `legacy-explorer` for consistency.

measurements/api/legacy_explorer.py
```diff
@@ -0,0 +1,135 @@
...
+## Something that has data:
+## https://explorer.ooni.torproject.org/api/reports/vendors?probe_cc=DE
+
+GOOD_REPORTS_ORDER = set('{} {}'.format(*t) for t in itertools.product(
+        ('input', 'probe_asn', 'probe_cc', 'test_name', 'test_start_time'),
+        ('ASC', 'DESC')))
+
+## {"test_start_time":{"between":["",""]}}
+## {"test_start_time":{"between":["",""]}}
+@api_legacy_explorer_blueprint.route('/reports')
```
hellais: Everything that is requested via the `/reports$` endpoints is already supported almost entirely (except for the link to the measurement) inside of this branch: https://github.com/TheTorProject/ooni-explorer/pull/119 that requires only ooni-explorer changes and works by querying the public endpoints.

I would suggest to maybe focus on implementing the the other endpoints (no need to have them be exactly 1-1 compatible, but as long as the needed data is exposed I will wire them up to explorer).

The wiring up to explorer is really not a lot of work, if you consider that to replace the `/reports` endpoint with the new measurements API it took about 2h of work. Now that I have figured out how to do it, doing the rest is going to be pretty simple.

measurements/api/legacy_explorer.py
darkk: I've put it under separate prefix for nginx & separate blueprint to `git rm` single file when explorer becomes "revamped". What's the drawback of this way?

darkk:
Cherry-picked valuable stuff, dropped non-valuable.

PR #45, opened 2017-12-05
Fix/download urls

hellais:
This fixes: https://github.com/TheTorProject/ooni-measurements/issues/44

Depends on: https://github.com/TheTorProject/ooni-measurements/pull/43

### DIFF
--- a/measurements/pages/__init__.py
+++ b/measurements/pages/__init__.py
@@ -24,6 +24,8 @@ pages_blueprint = Blueprint('pages', 'measurements',
                             static_url_path='/static/')
 
 
+DAY_REGEXP = re.compile("^\d{4}\-[0-1]\d\-[0-3]\d$")
+
 def _latest_reports():
     q = current_app.db_session.query(Report) \
             .order_by("textname DESC") \
@@ -142,11 +144,14 @@ def _files_on_date(date, order_by, order):
 
 @pages_blueprint.route('/files/by_date/<date>')
 def files_on_date(date):
-    # XXX do some validation of date
+    if not DAY_REGEXP.match(date):
+        raise BadRequest('Invalid date format')
+
     order_by = request.args.get('order_by', 'test_start_time')
     order = request.args.get('order', 'desc')
     if order.lower() not in ('desc', 'asc'):
-        raise BadRequest()
+        raise BadRequest('order must be desc or asc')
+
     if order_by not in ('test_start_time', 'probe_cc', 'report_id',
                         'test_name', 'probe_asn'):
         raise BadRequest()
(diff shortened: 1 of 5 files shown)


measurements/pages/__init__.py
```diff
@@ -275,6 +275,18 @@ def generator():
 
 @pages_blueprint.route('/files/download/<path:textname>')
 def files_download(textname):
+    if len(textname.split('/')) == 1:
```
darkk: `'/' not in textname` is maybe a bit more idiomatic.

measurements/pages/__init__.py
```diff
@@ -275,6 +275,18 @@ def generator():
...
+        # This is for backward compatibility with the new pipeline.
+        # See: https://github.com/TheTorProject/ooni-measurements/issues/44
+        q = current_app.db_session.query(Report.textname)\
+                .filter(Report.textname.endswith(textname))
```
darkk: Does `/files/download/.json` lead to slow query following an OOM?
Should it be `'/' + textname`?
Should `.limit(2)` be added?
Should exception be raised on two rows returned? IIRC, there were some reports that that same name in two buckets. Probably, these reports are duplicated, but I don't remember by heart.

measurements/pages/__init__.py
```diff
@@ -275,6 +275,18 @@ def generator():
...
+        first = q.first()
+        if first is None:
+            raise NotFound("No file with that filename found")
+
+        return redirect('/files/download/%s' % first[0])
+
+    textname_filter = (Report.textname == textname)
```
darkk: Is it debugging line? I don't see if it's used variable...

hellais: Ooops, yes should delete it.

measurements/pages/__init__.py
hellais: It does lead to pretty slow queries, but they are not that bad (on very slow network it's about 5-6 second)

measurements/pages/__init__.py
hellais: OK

measurements/pages/__init__.py
```diff
@@ -275,7 +275,7 @@ def generator():
 
 @pages_blueprint.route('/files/download/<path:textname>')
 def files_download(textname):
-    if len(textname.split('/')) == 1:
+    if '/' not in textname == 1:
```
darkk: That's typo, isn't it?

measurements/pages/__init__.py
darkk: How much RAM, CPU and network transfer does suggested pathological case take?
What query does it actually generate? Does `q.first()` implicitly add `LIMIT 1` or not?

hellais: > first() applies a limit of one within the generated SQL, so that only one primary entity row is generated on the server side
http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.first

PR #46, opened 2017-12-05
Feat/unbreak explorer

hellais:
This is more stuff on the API side of things to implement what is in: https://github.com/TheTorProject/ooni-explorer/pull/119


measurements/api/private.py
```diff
@@ -7,16 +7,21 @@
 """
 from datetime import datetime
 
+from six.moves.urllib.parse import urljoin
```
darkk: IMHO, `six` is useless as `ooni-measurement` is py3-only

hellais: Fair enough, but since we import and use it in other places already, I would rather do the removal of `six` globally than just here.

PR #47, opened 2017-12-13
Release/1.0.3

hellais:


measurements/__init__.py
```diff
@@ -2,4 +2,4 @@
 __email__ = "contact@openobservatory.org"
 
 __license__ = "BSD 2 Clause"
-__version__ = "1.0.2"
+__version__ = "1.0.3"
```
darkk: And package.json says `1.0.2`. IMHO these versions, releases and changelogs are a bit cargo-cultish here :)

PR #48, opened 2018-01-26
Use Request-Id header, forward it via HTTP and to postgres 

darkk:
That's already deployed as `20180126-9610c67d`.

@hellais, can you follow same pattern in new ooni-explorer, and old one, if that's trivial? I tried to do it myself for axios, but got a bit lost while understanding request context accessible from `apiClient.interceptors.request`.

### DIFF
--- a/measurements/config.py
+++ b/measurements/config.py
@@ -14,6 +14,7 @@ CACHE_CONFIG = {'CACHE_TYPE': 'simple'}
 
 APP_ENV = os.environ.get("APP_ENV", "development")
 DATABASE_URL = os.environ.get("DATABASE_URL", "postgresql://postgres@localhost:5432/ooni_measurements")
+DATABASE_STATEMENT_TIMEOUT = int(os.environ.get("DATABASE_STATEMENT_TIMEOUT", "0")) # to kill long-running statements ASAP
 
 BASE_URL = os.environ.get("BASE_URL", "https://api.ooni.io/")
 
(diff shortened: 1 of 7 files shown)


measurements/config.py
```diff
@@ -29,3 +32,10 @@
 
 # As of 2017-07-18 635830 is the latest index in the database
 REPORT_INDEX_OFFSET = int(os.environ.get("REPORT_INDEX_OFFSET", "635830"))
+
+REQID_HDR = "Request-Id"
```
hellais: You should probably use some more generic header to avoid possible clashes. For example `X-OONI-Request-Id`

darkk: [Generic](https://devcenter.heroku.com/articles/http-request-id) one is `X-Request-Id`, but it's not generic enough as noone had bothered to [register it with IANA](http://www.iana.org/assignments/message-headers/message-headers.xhtml). :-) Also, [`X-` prefix is deprecated by RFC6648](https://tools.ietf.org/html/rfc6648).

Anyway, I intentionally use generic one and [filter it on ingress](https://github.com/TheTorProject/ooni-sysadmin/blob/6e65c46af2ec3f742e87b167566ee6fc5381d723/ansible/roles/nginx-stable/templates/nginx.conf#L17-L33) as it has to be filtered anyway.

darkk: But that's good catch, I'd rather use `X-Request-ID` that is even more generic.

PR #58, opened 2018-09-10
Update the charts on the stats page

darkk:
* Only show the last 24 months
* Don't show the unfinalized month
* Round the last day of the month
* Add marker for OONI Probe mobile app launch

@hellais is it good to merge and deploy?


hellais:
I took a quick look at the diff and it looks good to me.

Thanks for putting this together 👍

PR #64, opened 2019-01-09
Fix travis build

hellais:
* pip freeze dependencies to last known working version
* Add .python-version and remove unneeded travis command


.travis.yml
```diff
@@ -35,8 +35,6 @@ before_script:
 script:
 - node -v
 - npm -v
-# Build the docker image
-- make APP_ENV=testing build
```
hellais: I dropped this because it doesn't actually exist in the Makefile (anymore?)

Makefile
```diff
@@ -35,7 +35,7 @@ shell:
 	$(PYTHON_WITH_ENV) -m measurements shell
 
 create-tables:
-	$(PYTHON_WITH_ENV) -m measurements create_tables
+	$(PYTHON_WITH_ENV) -m measurements create-tables
```
hellais: In click versions 7+ the command is transformed into `dash-case`

PR #70, opened 2019-01-13
Better query timeout exception capturing

hellais:
Also collect the sentry sql_query context

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -10,6 +10,8 @@ import connexion
 import requests
 import lz4framed
 
+from sentry_sdk import configure_scope
+
 from flask import Blueprint, current_app, request, make_response
 from flask.json import jsonify
 from werkzeug.exceptions import HTTPException, BadRequest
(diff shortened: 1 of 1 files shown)


measurements/api/measurements.py
```diff
@@ -323,67 +332,71 @@ def list_measurements(
     iter_start_time = time.time()
     results = []
 
-    try:
-        for row in q:
-            measurement_id = '{}-{}'.format(MSM_ID_PREFIX, row.msm_no)
-            url = urljoin(
+    with configure_scope() as scope:
+        scope.set_extra("sql_query", str(q))
```
hellais: This will log next to the error the `sql_query` which triggered the error, to make debugging easier. See: https://docs.sentry.io/enriching-error-data/context/?platform=python

measurements/api/measurements.py
```diff
@@ -323,67 +332,71 @@ def list_measurements(
...
+                    'input': row.input,
+                    'anomaly': row.anomaly,
+                    'confirmed': row.confirmed,
+                    'failure': (row.exc != None
+                                or row.residual_no != None
+                                or row.msm_failure),
+                })
+        except exc.OperationalError as exc:
+            if isinstance(exc.orig, QueryCanceledError):
+                raise QueryTimeoutError()
```
hellais: I changed the error capturing to be a bit more precise (we actually should trap the sqlalchemy sub exception) and made a custom exception to make filtering in the sentry UI easier.

PR #74, opened 2019-01-21
Capture timeouts

hellais:
Now that timeouts are handled properly, they are not logged to sentry anymore.

This PR is about capturing the timeouts explicitly if we believe that is a useful thing to do.

We do run the risk of spamming sentry with too many exceptions, but maybe it's useful for us to optimise the database.


hellais:
I am self merging this as I think it's a reasonable thing to do.

PR #76, opened 2019-02-27
Fix/75

hellais:
This fixes the bug in the stats page and also upgrades some old dependencies.


hellais:
I am merging this to make a new release.

PR #77, opened 2019-03-13
Implement views for OONI Explorer beta

hellais:
These are the views that are necessary to implementing the public beta of the next generation OONI Explorer.

These views are mostly related to the OONI Explorer country page:

* [x] `/api/_/test_coverage` provides stats on the number of measurements per test group (and the network coverage) over the last 30 days
* [x] `/api/_/website_urls` returns the list of URLs that have been tested in a given country & network pair
* [x] `/api/_/website_networks` return the list of networks covered by websites tests in a given country
* [x] `/api/_/website_stats`  return the 30 day counts for a given input, probe_cc, probe_asn of anomalies, failures, confirmed and total count
* [x] `/api/_/im_networks` shall return the list of networks covered by IM tests grouped by im app name
* [x] `/api/_/im_stats` shall return the count of measurements given a `probe_asn`, `test_name` & `probe_cc`
* [x] `/api/_/network_stats` shall return stats related to performance and middleboxes being detected grouped by the network
* [x] `/api/_/country_overview` provides a high level overview of what is known WRT to censorship in a particular country

Other items that need to be done include:

* [x] Returning the URL count in the website_urls listing
* [x] Fill the "holes" in the charts looking at days
* [x] Better pagination of the websites tested section
* [x] Enriching the networks related endpoints with the names of networks
* ~[ ] Enriching the URLs related endpoints with category codes~
* ~[ ] Add support for searching for tested URLs in the website_urls section~

These views are supported by a series of fact tables that should be incrementally updated by the pipeline once the batch process ticks.
The tables currently being used are: `ooexpl_recent_msmt_count`, `ooexpl_bucket_msmt_count`, `ooexpl_website_msmts`.

In terms of rolling this out I would image we could go about doing it as follows:

* [x] Roll this out to the staging api.test.ooni.io with views being created and populated as UNLOGGED tables
* [x] Add support for running the incremental update of these tables as part of the data processing pipeline stage/workflow (xref: https://github.com/ooni/pipeline/pull/144)
* [x] Roll this out on the production api.ooni.io when we are ready to launch the beta of OONI Explorer

It would especially to be useful to have some guidance by @darkk on the approach I am taking with respect to incrementally updating these tables and deploying this on the pipeline in a way that it can support a production OONI Explorer.

I think this is also a good opportunity to refactor the current legacy OONI Explorer endpoints to use these fact tables instead of relying on the static dumps which have not been updated in more than a year at this point (see: https://github.com/ooni/api/blob/master/measurements/api/private.py#L221).
From what I can tell it should be pretty easy to adapt the shape of the above endpoints to fit the needs of the endpoints.


hellais:
@sarathms With respect to implementation the workflow you should be following for implementing the websites tests chart section is as follows:

1. You retrieve the list of networks by calling: https://api.test.ooni.io/api/_/website_networks?probe_cc=MY
With this you will be drawing the listing of networks for which we have measurements and autocomplete the selector to have the first item in the list (they are already sorted by the most relevant network)

2. You retrieve the list of websites by calling https://api.test.ooni.io/api/_/website_urls?probe_cc=MY&probe_asn=4788, the value of `probe_asn` will come from the first item in the list in 1.
With this you will be populating the first page of the network listing. Note: you don’t yet have the actual data points for the chart, but only the listing of URLs

3. For each website in the above list (they should be 10, though I now see I made a mistake in the code which I will fix) you start an async request for: https://api.test.ooni.io/api/_/website_stats?probe_cc=MY&probe_asn=4788&input=http://thepiratebay.se
As soon as the responses come back you will then be able to draw the charts in the websites sections

I tried to imagine how I would be using the API if I were to be coding the react components for this page and tried to make it easy to use for the use-case of the UI, but if you would like me to tweak the API or workflow to work better with what you had in mind, let me know and I can probably come up with something.

Note: similarly to the other API endpoints, these may have some holes (if there are no measurements on a certain day you will not get a datapoint with zero in it). I can fix that if it proves to be a challenge to reshape it in JS

darkk:
FYI I've noticed the bug in, probably(!), this branch as [the bump references non-existent commit](https://github.com/ooni/sysadmin/pull/286/commits/fa6e5f2d900999ef467fa17351001c909b25388b) but [container is pushed](https://hub.docker.com/r/openobservatory/api/tags) and it probably is commit 07210b95. I assume that as the previous container refers the latest commit in this branch, namely https://github.com/ooni/api/pull/77/commits/d3d2ccd75cea94e0f85f8003fa5faaa40aa31de3

`WHERE date_trunc('day', report.test_start_time) = '2017-05-05'` is a full-scan of 2Gb table:
```
 Sort  (cost=355011.48..355092.59 rows=32443 width=277) (actual time=3410.463..3411.713 rows=1785 loops=1)
   Sort Key: test_start_time DESC
   Sort Method: quicksort  Memory: 969kB
   ->  Seq Scan on report  (cost=0.00..348366.59 rows=32443 width=277) (actual time=607.873..3408.115 rows=1785 loops=1)
         Filter: (date_trunc('day'::text, test_start_time) = '2017-05-05 00:00:00'::timestamp without time zone)
         Rows Removed by Filter: 6486257
 Planning time: 0.161 ms
 Execution time: 3412.901 ms
```
Re-phrasing it as `report.test_start_time >= '2017-05-05' and report.test_start_time < '2017-05-06'` is significantly different:
```
 Sort  (cost=8378.06..8383.57 rows=2206 width=277) (actual time=38.216..39.852 rows=1785 loops=1)
   Sort Key: test_start_time DESC
   Sort Method: quicksort  Memory: 969kB
   ->  Bitmap Heap Scan on report  (cost=51.04..8255.55 rows=2206 width=277) (actual time=30.927..34.782 rows=1785 loops=1)
         Recheck Cond: ((test_start_time >= '2017-05-05 00:00:00'::timestamp without time zone) AND (test_start_time < '2017-05-06 00:00:00'::timestamp without time zone))
         Heap Blocks: exact=100
         ->  Bitmap Index Scan on report_test_start_time_idx  (cost=0.00..50.49 rows=2206 width=0) (actual time=30.874..30.874 rows=1785 loops=1)
               Index Cond: ((test_start_time >= '2017-05-05 00:00:00'::timestamp without time zone) AND (test_start_time < '2017-05-06 00:00:00'::timestamp without time zone))
 Planning time: 0.187 ms
 Execution time: 41.460 ms
```

hellais:
@darkk where did you find this query and those runtimes?

I can't seem to find a reference to that query in the api or pipeline code.

The closest query I can think of is this one: https://github.com/ooni/pipeline/blob/ooexplr-meta/af/oometa/014-ooexpl-insert.install.sql#L22, but it's not doing a comparison on the truncated start_time.

darkk:
Ah, sorry, indeed I was wrong that it was relevant to *this* branch, I was confused by the container referencing non-pushed commit.

> I can't seem to find a reference to that query in the api or pipeline code.

It looks like the query is coming from [`_files_on_date()`](https://github.com/ooni/api/blob/9610c67/measurements/pages/__init__.py#L139-L143).

> where did you find this query and those runtimes

I was waiting for `UPDATE anomaly` to converge and polled `pg_stat_activity`. The slow query from API at `hkgmetadb` attracted my attention as it was slow without good reason to be slow. I assumed that beta API instance is usually firing queries relevant to the "beta" part and the branch had some non-pushed commits, so I commented to this branch instead of filing a separate ticket.

hellais:
> It looks like the query is coming from _files_on_date().

I filed a separate ticket for this here: https://github.com/ooni/api/issues/83.

> I was waiting for UPDATE anomaly to converge and polled pg_stat_activity

Speaking of that, the UPDATE anomaly query is still running after almost 24h:

```
3426	23:45:20.896896	shovel	BEGIN;  UPDATE measurement SET anomaly = TRUE, confirmed = TRUE WHERE msm_no IN (SELECT msm_no FROM http_request_fp); UPDATE measurement SET anomaly = TRUE WHERE msm_no IN (SELECT msm_no FROM http_verdict WHERE blocking != 'false' AND blocking IS NOT NULL);  COMMIT;	active
```

Do you think this is normal or to be expected?

darkk:
> Do you think this is normal or to be expected?

I'm unsure what you mean asking if it's normal. It sounds like a really aggressive question to me, so I'm unsure if I understand it correctly.

Speaking of the expectations, I'd say it's reasonably expected runtime from the PostgreSQL query planner point of view. E.g. the second query estimates query cost as 5459432.74 .. 24260758.59. As far as I understand, 4 units of cost estimates the time for one random-read IO operation. So the cost is roughly estimated as 0.3...1.2 days with 60 IOPS (coming from random read taking ~15ms per `seeksample.py` stats in [prometheus](https://mon.ooni.nu/prometheus/graph?g0.range_input=7d&g0.expr=rate(node_seeksample_seconds_sum%7Bdevice%3D%22xvda1%22%2Cinstance%3D%22hkgmetadb.infra.ooni.io%3A9100%22%2Cjob%3D%22node%22%7D%5B10m%5D)%20%2F%20rate(node_seeksample_seconds_count%5B10m%5D)&g0.tab=0)). That estimate is likely inaccurate, but I expect the error to be less than an order of magnitude. So it does not sound like something unexpected.

hellais:
> As far as I understand, 4 units of cost estimates the time for one random-read IO operation. So the cost is roughly estimated as 0.3...1.2 days with 60 IOPS (coming from random read taking ~15ms per seeksample.py stats in prometheus)

Ah thanks for explaining this. 

I guess what I meant with the questions is if you think I did something very silly in forming the query (I made some changes to it compared to how it was in the migration scripts) and if you think there is reason to worry about it not converging and/or if it makes sense to keep waiting.

Based on the stats your provide it may be the case that it will come to a conclusion within the next hours hopefully.

PR #88, opened 2019-05-03
Performance

hellais:
Some improvements to measure performance of the API better and possibly boost it.

### DIFF
--- a/measurements/gunicorn_config.py
+++ b/measurements/gunicorn_config.py
@@ -1,5 +1,7 @@
 from .config import PROMETHEUS_PORT
 
+# This config is taken from:
+# https://github.com/rycus86/prometheus_flask_exporter#multiprocess-applications
 from prometheus_flask_exporter.multiprocess import GunicornPrometheusMetrics
 
 def when_ready(server):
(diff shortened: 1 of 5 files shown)


measurements/gunicorn_config.py
```diff
@@ -0,0 +1,9 @@
+from .config import PROMETHEUS_PORT
```
bassosimone: I'd probably mention that this is the vanilla usage according to the docs

measurements/wsgi.py
```diff
@@ -1,6 +1,12 @@
...
 from werkzeug.contrib.fixers import ProxyFix
 
 from measurements.app import create_app
 
+from prometheus_flask_exporter.multiprocess import GunicornPrometheusMetrics
+
 application = create_app()
 application.wsgi_app = ProxyFix(application.wsgi_app)
+
+metrics = GunicornPrometheusMetrics(application, group_by='endpoint')
```
bassosimone: I assume you will use `metrics` later?

hellais: There is stuff that can be done by using metrics, but I don't need to currently. I put it inside of a variable so we don't forget that this instance can be used to further customise the metrics endpoint.

measurements/gunicorn_config.py
hellais: What do you mean by this?

bassosimone: That this code is exactly the default way of deploying if you read the docs.

PR #89, opened 2019-05-12
Readme: add building assets

pellaeon:


hellais:
yes this is reasonable

PR #103, opened 2019-08-30
Use the materialized view ooexpl_wc_input_counts

hellais:
This is related to: https://github.com/ooni/pipeline/pull/222


hellais:
The CI is failing because the DB initialization logic is not yet merged in ooni/pipeline#222.

PR #104, opened 2019-09-03
Optimize the api_private_stats endpoints

hellais:
This changes a little bit the meaning of the stats queries as it considers the time in which a measurement is uploaded instead of the time in which it was run.

I think this is pretty consistent with what we present in the UIs and it's significantly faster to do it this way so I think it's ok.

The performance gain for these queries is in the order of 10-100x.

### DIFF
--- a/measurements/api/private.py
+++ b/measurements/api/private.py
@@ -918,4 +918,7 @@ def api_private_global_overview():
         'country_count': row[0],
         'network_count': row[1],
         'measurement_count': row[2],
+        'networks_by_month': api_private_stats_by_month('COUNT(DISTINCT probe_asn)'),
+        'countries_by_month': api_private_stats_by_month('COUNT(DISTINCT probe_cc)'),
+        'measurements_by_month': api_private_stats_by_month('SUM(count)')
     })
--- a/measurements/api/private.py
+++ b/measurements/api/private.py
@@ -59,15 +59,15 @@ def api_private_stats_by_month(orm_stat):
         'date': (bkt + relativedelta(months=+1, days=-1)).strftime("%Y-%m-%d"),
         'value': value,
     } for bkt, value in sorted(r)]
-    return jsonify(result)
+    return result
 
 @api_private_blueprint.route('/asn_by_month')
 def api_private_asn_by_month():
-    return api_private_stats_by_month('COUNT(DISTINCT probe_asn)')
+    return jsonify(api_private_stats_by_month('COUNT(DISTINCT probe_asn)'))
 
 @api_private_blueprint.route('/countries_by_month')
 def api_private_countries_by_month():
-    return api_private_stats_by_month('COUNT(DISTINCT probe_cc)')
+    return jsonify(api_private_stats_by_month('COUNT(DISTINCT probe_cc)'))
 
 @api_private_blueprint.route('/runs_by_month')
 def api_private_runs_by_month():
--- a/measurements/api/private.py
+++ b/measurements/api/private.py
@@ -917,7 +917,12 @@ def api_private_global_overview():
     return jsonify({
         'country_count': row[0],
         'network_count': row[1],
-        'measurement_count': row[2],
+        'measurement_count': row[2]
+    })
+
+@api_private_blueprint.route('/global_overview_by_month', methods=["GET"])
+def api_private_global_by_month():
+    return jsonify({
         'networks_by_month': api_private_stats_by_month('COUNT(DISTINCT probe_asn)'),
         'countries_by_month': api_private_stats_by_month('COUNT(DISTINCT probe_cc)'),
         'measurements_by_month': api_private_stats_by_month('SUM(count)')
--- a/measurements/api/private.py
+++ b/measurements/api/private.py
@@ -40,16 +40,21 @@ def api_private_stats_by_month(orm_stat):
     # implemented in a better way, but that sanity check is probably enough.
     ## ooni_epoch = datetime(2012, 12, 1)
 
+    now = datetime.now()
+    end_date = datetime(now.year, now.month, 1)
+    start_date = end_date - relativedelta(months=24)
+
     s = select([
         func.date_trunc('month', sql.text('bucket_date')).label('bucket_month'),
-        orm_stat
-    ]).where(
-        sql.text("bucket_date > current_date - interval '24 month'")
-    ).group_by(
+        sql.text(orm_stat)
+    ]).where(and_(
+        sql.text("bucket_date > :start_date"),
+        sql.text("bucket_date < :end_date")
+    )).group_by(
         'bucket_month'
     ).select_from(sql.table('ooexpl_bucket_msm_count'))
 
-    r = current_app.db_session.execute(s)
+    r = current_app.db_session.execute(s, {'start_date': start_date, 'end_date': end_date})
     result = [{
         'date': (bkt + relativedelta(months=+1, days=-1)).strftime("%Y-%m-%d"),
         'value': value,
--- a/measurements/api/private.py
+++ b/measurements/api/private.py
@@ -63,21 +63,30 @@ def api_private_stats_by_month(orm_stat):
 
 @api_private_blueprint.route('/asn_by_month')
 def api_private_asn_by_month():
-    # The query takes ~6s on local SSD @ AMS on 2018-04-04.
-    # It was taking ~45s when it was fetching all the table from DB and doing grouping locally.
     return api_private_stats_by_month('COUNT(DISTINCT probe_asn)')
 
 @api_private_blueprint.route('/countries_by_month')
 def api_private_countries_by_month():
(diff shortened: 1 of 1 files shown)


hellais:
I did some tests to see how different the numbers are between the old and the new method.

Here are the results:
`print_diff(old_asns_by_month, new_asns_by_month)`
```
2017-09-30    -7.85%% (-133)
2017-10-31    1.04%% (18)
2017-11-30    -0.03%% (-1)
2017-12-31    0.50%% (23)
2018-01-31    -0.88%% (-50)
2018-02-28    -0.44%% (-22)
2018-03-31    -0.26%% (-13)
2018-04-30    -0.20%% (-10)
2018-05-31    -0.33%% (-16)
2018-06-30    -0.41%% (-19)
2018-07-31    0.47%% (22)
2018-08-31    -0.20%% (-9)
2018-09-30    -0.88%% (-41)
2018-10-31    0.43%% (22)
2018-11-30    -2.85%% (-135)
2018-12-31    0.09%% (4)
2019-01-31    1.06%% (43)
2019-02-28    -0.57%% (-26)
2019-03-31    0.00%% (0)
2019-04-30    1.58%% (76)
2019-05-31    -1.10%% (-57)
2019-06-30    -0.83%% (-42)
2019-07-31    -0.20%% (-10)
2019-08-31    -0.13%% (-7)
```
`print_diff(old_countries_by_month, new_countries_by_month)`
```
2017-09-30    -3.92%% (-6)
2017-10-31    0.65%% (1)
2017-11-30    0.00%% (0)
2017-12-31    0.00%% (0)
2018-01-31    0.00%% (0)
2018-02-28    0.47%% (1)
2018-03-31    -0.47%% (-1)
2018-04-30    0.00%% (0)
2018-05-31    0.48%% (1)
2018-06-30    0.00%% (0)
2018-07-31    0.48%% (1)
2018-08-31    -0.48%% (-1)
2018-09-30    0.00%% (0)
2018-10-31    0.00%% (0)
2018-11-30    0.00%% (0)
2018-12-31    0.00%% (0)
2019-01-31    0.00%% (0)
2019-02-28    0.00%% (0)
2019-03-31    0.00%% (0)
2019-04-30    0.00%% (0)
2019-05-31    -0.47%% (-1)
2019-06-30    0.00%% (0)
2019-07-31    0.00%% (0)
2019-08-31    0.00%% (0)
```

`print_diff(old_runs_by_month, new_runs_by_month)`
```
2017-09-30    -6444.16%% (-6226544)
2017-10-31    -5349.50%% (-7484641)
2017-11-30    -3892.42%% (-7292919)
2017-12-31    -2744.37%% (-7844582)
2018-01-31    -2201.24%% (-9380161)
2018-02-28    -2537.63%% (-7569984)
2018-03-31    -2789.88%% (-8475364)
2018-04-30    -3048.32%% (-8583728)
2018-05-31    -3150.51%% (-8215501)
2018-06-30    -3647.68%% (-8787080)
2018-07-31    -4155.93%% (-9917265)
2018-08-31    -3881.84%% (-9378534)
2018-09-30    -3437.50%% (-8898622)
2018-10-31    -2211.14%% (-8642072)
2018-11-30    -2484.88%% (-9075293)
2018-12-31    -2938.13%% (-8763370)
2019-01-31    -3746.96%% (-9005507)
2019-02-28    -3049.89%% (-8006477)
2019-03-31    -1238.15%% (-8703950)
2019-04-30    -2497.57%% (-8040347)
2019-05-31    -2775.31%% (-9939757)
2019-06-30    -2350.09%% (-8298167)
2019-07-31    -2377.25%% (-8479707)
2019-08-31    -2524.54%% (-9478752)
```

So it seems like for countries and ASN the difference is minimal, for the runs however the difference is huge.

I think the reason for this is that previously we were counting the number of reports in a time period, while now we are counting the number of measurements, hence the count is going to be significantly larger.

Maybe what we can do is preserve the legacy behaviour for the runs, but adjust OONI Explorer to display the measurements in a month instead of the runs.

hellais:
So I have ended up restoring the behaviour of the runs_by_month endpoints and moved the measurement_by_month count into the `global_overview` API endpoint.

@sarathms is it hard to change how explorer gets data for the chart on the landing page to get it from `/api/_/global_overview` (which it's already using for the global stats in the landing page) instead of the `/api/_/runs_by_month`, `/api/_/networks_by_month` etc.?

sarathms:
@hellais It is not difficult to use `/api/_/global_overview` to fetch chart data. The biggest difference is that this call happens on the server side, so it can affect the perceived performance of the page. Specifically, that it might take longer to start rendering the page. Now, we render the page, make the API calls, re-render the section.

measurements/api/private.py
```diff
@@ -44,34 +44,49 @@ def api_private_stats_by_month(orm_stat):
...
         'value': value,
     } for bkt, value in sorted(r)]
-    return jsonify(result)
+    return result
 
 @api_private_blueprint.route('/asn_by_month')
 def api_private_asn_by_month():
-    # The query takes ~6s on local SSD @ AMS on 2018-04-04.
-    # It was taking ~45s when it was fetching all the table from DB and doing grouping locally.
-    return api_private_stats_by_month( func.count(distinct(Report.probe_asn)) )
```
bassosimone: I see that you are removing these comments but cannot guess why you're doing it now

measurements/api/private.py
```diff
@@ -44,34 +44,49 @@ def api_private_stats_by_month(orm_stat):
...
+    return jsonify(api_private_stats_by_month('COUNT(DISTINCT probe_asn)'))
 
 @api_private_blueprint.route('/countries_by_month')
 def api_private_countries_by_month():
-    # The query takes ~10s on local SSD @ AMS on 2018-04-04.
-    # It was taking ~25s when it was fetching all the table from DB and doing grouping locally.
```
bassosimone: Ditto

measurements/api/private.py
```diff
@@ -44,34 +44,49 @@ def api_private_stats_by_month(orm_stat):
...
-        ).group_by('test_start_month')
+    s = select([
+        func.date_trunc('month', sql.text('bucket_date')).label('bucket_month'),
+        sql.text(orm_stat)
+    ]).where(and_(
+        sql.text("bucket_date > :start_date"),
+        sql.text("bucket_date < :end_date")
+    )).group_by(
+        'bucket_month'
+    ).select_from(sql.table('ooexpl_bucket_msm_count'))
```
bassosimone: This seems to me another way to prepare the query. What is the reason? Why is it faster?

measurements/api/private.py
```diff
@@ -44,34 +44,49 @@ def api_private_stats_by_month(orm_stat):
...
+
+    r = current_app.db_session.query(
+            func.date_trunc('month', Report.test_start_time).label('test_start_month'),
+            func.count()
+        ).filter(Report.test_start_time >= start_date).filter(Report.test_start_time < end_date
+        ).group_by('test_start_month')
+    result = [{
+        'date': (bkt + relativedelta(months=+1, days=-1)).strftime("%Y-%m-%d"),
+        'value': value,
+    } for bkt, value in sorted(r)]
```
bassosimone: I'm failing to understand the rationale of this change, can you explain?

measurements/api/private.py
bassosimone: (I have seen some benchmarks above, but I'm still a bit in the dark.)

hellais: The query is now using a different table which has some pre-computed stats inside of it. Since the stats in that table are precomputed it's now significantly faster.

measurements/api/private.py
hellais: Because the query time is no longer so slow (I have done optimisations on it) so the comment is no longer relevant.

measurements/api/private.py
hellais: Same as above.

measurements/api/private.py
hellais: This is basically just restoring the code for this particular API endpoint to how it was before.

See: https://github.com/ooni/api/pull/104#issuecomment-527965586 & https://github.com/ooni/api/pull/104#issuecomment-528823688 for a more in-depth explaination as to why we want to preseve the old slow query for this endpoint vs using the materialized table for the others (tl;dr the materialized tables don't have enough information in them to be the same result).

hellais:
> The biggest difference is that this call happens on the server side, so it can affect the perceived performance of the page. Specifically, that it might take longer to start rendering the page. Now, we render the page, make the API calls, re-render the section

That is a very good point. I am going to move these into a separate endpoint then, but still as part of a single API call.

sarathms:
> but still as part of a single API call

And that should already make it better than what we have now.

measurements/api/private.py
hellais: > What was the table before?

`report`. This is implied by the fact that in the query you are using the SQLAlchemy ORM `Report` object.

bassosimone: Ah, I see, thanks for explaining

PR #105, opened 2019-09-05
Faster domain filter

hellais:
Query time for an example query has gone from 537,136.721 ms
to 146.392 ms.

This improvement is achieved by creating a CTE and using it to hint the
query planner as to how to do the cross table JOIN.

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -16,7 +16,8 @@ from flask import Blueprint, current_app, request, make_response
 from flask.json import jsonify
 from werkzeug.exceptions import HTTPException, BadRequest
 
-from sqlalchemy import func, or_, and_, false, true, text
+from sqlalchemy.dialects import postgresql
+from sqlalchemy import func, or_, and_, false, true, text, select, sql, column
 from sqlalchemy.orm import lazyload
 from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 from sqlalchemy.exc import OperationalError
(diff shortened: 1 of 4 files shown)


measurements/api/measurements.py
```diff
@@ -16,7 +16,8 @@
 from flask.json import jsonify
 from werkzeug.exceptions import HTTPException, BadRequest
 
-from sqlalchemy import func, or_, and_, false, true, text
+from sqlalchemy.dialects import postgresql
+from sqlalchemy import func, or_, and_, false, true, text, select, sql, column
```
FedericoCeratto: "true" is unused

measurements/database.py
```diff
@@ -9,6 +9,7 @@
 from sqlalchemy import event
 from sqlalchemy.engine import Engine
 
+from sqlalchemy.dialects import postgresql
```
FedericoCeratto: unused

measurements/api/measurements.py
```diff
@@ -200,31 +201,33 @@ def get_measurement(measurement_id, download=None):
                 'attachment', filename=filename)
     return response
 
-def input_filter(q, input_, domain, test_name):
+def input_cte(input_, domain, test_name):
```
FedericoCeratto: input_cte appears not to be covered in https://travis-ci.org/ooni/api/builds/581319654#L2112  - is it worth testing?

hellais: I added a basic check to the smoke test. Without fixtures in the DB though it's hard to have a really useful test in here though.

PR #115, opened 2019-09-19
Implement Fastpath support in the API for faster publication of measurements

FedericoCeratto:
Add integration tests

Related to https://github.com/ooni/pipeline/pull/213 and https://github.com/ooni/api/issues/72


measurements/api/measurements.py
```diff
@@ -214,21 +250,42 @@ def input_cte(input_, domain, test_name):
             text('input.input LIKE :i').bindparams(i='%{}%'.format(input_)),
         )
 
-    if domain:
```
FedericoCeratto: Apart from the "if" this is just a reflow.

measurements/api/measurements.py
```diff
@@ -361,26 +423,83 @@ def list_measurements(
...
     iter_start_time = time.time()
     results = []
 
-    query_str = str(q.statement.compile(dialect=postgresql.dialect()))
     with configure_scope() as scope:
         scope.set_extra("sql_query", query_str)
 
         try:
             for row in q:
-                measurement_id = '{}-{}'.format(MSM_ID_PREFIX, row.msm_no)
```
FedericoCeratto: Moved to coalesce() above

measurements/models.py
```diff
@@ -12,79 +12,91 @@
 from sqlalchemy.orm import relationship
 
 from sqlalchemy import (
```
FedericoCeratto: Reflow

measurements/models.py
```diff
@@ -12,79 +12,91 @@
...
+    Boolean,
+    ForeignKey,
 )
 
-from sqlalchemy.dialects.postgresql import (
-    INET, ENUM, BYTEA, UUID
-)
+from sqlalchemy.dialects.postgresql import INET, ENUM, BYTEA, UUID
 
 TEST_NAMES = {
```
FedericoCeratto: Reflow

measurements/models.py
```diff
@@ -137,22 +151,38 @@ class Measurement(Base):
...
     test_name = Column(String)
     test_version = Column(String)
     software_name = Column(String)
     software_version = Column(String)
 
-    reports = relationship('Report', back_populates='software')
+    reports = relationship("Report", back_populates="software")
+
+
+class Fastpath(Base):
```
FedericoCeratto: This is new

measurements/api/measurements.py
```diff
@@ -27,9 +27,10 @@
 
 from measurements import __version__
 from measurements.config import REPORT_INDEX_OFFSET, REQID_HDR, request_id
-from measurements.models import Report, Input, Measurement, Autoclaved
+from measurements.models import Report, Input, Measurement, Autoclaved, Fastpath
 
 MSM_ID_PREFIX = 'temp-id'
+FASTPATH_MSM_ID_PREFIX = 'tid-'
```
hellais: Can we make this prefix more clearly temporary and related to the fastpath?

Maybe `temp-id-fast-`?

measurements/api/measurements.py
```diff
@@ -147,8 +148,40 @@ def list_files(
...
+        raise BadRequest("No measurement found")
+
+
+
 def get_measurement(measurement_id, download=None):
+    """Get one measurement by measurement_id,
+    fetching the file from S3 or the fastpath host as needed
+    Returns only the measurement without extra data from the database
+    """
     # XXX this query is SUPER slow
```
hellais: We should probably drop this comment as it's actually no longer true.

measurements/api/measurements.py
hellais: The idea was that you could pass in either of the two or possibly both and the filter would apply to both.

By dropping the `if` and making it into an `else` this function is no longer equivalent to the previous function.

I suggest putting it back to how it was and applying the reflow to the original function.

measurements/api/measurements.py
```diff
@@ -201,7 +234,10 @@ def get_measurement(measurement_id, download=None):
                 'attachment', filename=filename)
     return response
 
+
 def input_cte(input_, domain, test_name):
+    """Given a domain or an input_, build a WHERE filter
```
hellais: This should gave, given a `domain` and/or `input_`

measurements/api/measurements.py
```diff
@@ -201,7 +234,10 @@ def get_measurement(measurement_id, download=None):
...
+    """
     if input_ and domain:
         raise BadRequest("Must pick either domain or input")
```
hellais: So actually what I was saying before doesn't seem to be accurate I can see now from here.

I suppose then you can ignore my previous two comments as it actually only takes `input_` OR `domain`.

We probably don't actually want to support filtering by either of the two as it will likely lead to increased query time.

measurements/api/measurements.py
hellais: Ignore this last comment.

measurements/api/measurements.py
hellais: ignore this

measurements/api/measurements.py
```diff
@@ -361,26 +423,83 @@ def list_measurements(
...
+    if test_name is not None:
+        fpq = fpq.filter(Fastpath.test_name == test_name)
+    if input_:
+        fpq = fpq.filter(Fastpath.input == input_)
+
+    # Assemble the query
+
     q = q.limit(limit).offset(offset)
+    try:
+        q = q.union(fpq)
```
hellais: How do you ensure that a record is not duplicated inside of the fastpath and the batch pipeline and hence you will return duplicate rows by doing a UNION?

measurements/api/measurements.py
```diff
@@ -361,26 +423,83 @@ def list_measurements(
...
         q = q.order_by(text('{} {}'.format(order_by, order)))
 
-    query_time = 0
+
+    ## Create a query for fastpath
+
+    fpcols = [
+        func.coalesce(0).label('m_input_no'),
+        Fastpath.measurement_start_time.label("measurement_start_time"),
+        func.concat("tid-", Fastpath.tid).label('measurement_id'),
```
hellais: You should use the variable you have defined as `FASTPATH_MSM_ID_PREFIX` for this.

measurements/api/measurements.py
FedericoCeratto: I was trying to keep it short to keep the URL easier to handle and copypaste. temp-id-fast will match "temp-id" startswith() and might break something. `temp-fid` ?

hellais: > temp-id-fast will match "temp-id" startswith() and might break something. 
That's a good point.

`temp-fid` looks reasonable to me.

I mostly suggest to keep `temp` in the prefix so that it's clear that this ID is a temporary one and will likely change in the future.

hellais: > temp-id-fast will match "temp-id" startswith() and might break something.

That's a good point.

`temp-fid` looks reasonable to me.

I mostly suggest to keep temp in the prefix so that it's clear that this ID is a temporary one and will likely change in the future.

FedericoCeratto: Done

measurements/api/measurements.py
FedericoCeratto: The slowness? It seems to be using approx .5 seconds according to the /api/v1.measurements_api_get_measurement metric on Prometheus

measurements/api/measurements.py
FedericoCeratto: This was the previous behavior of the API, I just added a comment.

measurements/api/measurements.py
FedericoCeratto: Updated to return only one (merged) result

measurements/api/measurements.py
hellais: I believe in fact it used to be even slower, but maybe I am mistaken.

In any case perhaps it's useful to add some extra details about the slowness (and maybe link to a github issue) also in light of the fact that the comment is now above a different bit of code than it was originally.

FedericoCeratto: Updated.

measurements/api/measurements.py
```diff
@@ -214,21 +269,84 @@ def input_cte(input_, domain, test_name):
...
+    else:
+        if b["scores"]:
+            # merge in useful fields from fastpath
+            for f in ("scores", "measurement_url", "measurement_id"):
+                a[f] = b[f]
+            return a
+
+        else:
+            # both trad, ignore b
+            return a
```
hellais: Maybe we can improve a bit the readability of this function by reducing the level of nesting and explcitly listing all the cases, for example:
```
if a["scores"] and b["scores"]:
  # both fastpath, ignore b
  return a
elif a["scores"]:
  # merge in useful fields from traditional
  ...
elif b["scores"]:
  # merge in useful fields from fastpath
 ...
else:
  # both trad, ignore b
 ...      
```

What do you think?

measurements/api/measurements.py
```diff
@@ -147,8 +151,56 @@ def list_files(
...
+        "Incoming fastpath query %r. Fetching %s:%d%s",
+        measurement_id,
+        FASTPATH_SERVER,
+        FASTPATH_PORT,
+        path,
+    )
+    conn = http.client.HTTPConnection(FASTPATH_SERVER, FASTPATH_PORT)
+    log.debug("Fetching %s:%d %r", FASTPATH_SERVER, FASTPATH_PORT, path)
+    conn.request("GET", path)
+    r = conn.getresponse()
```
hellais: Is there a reason to use `http.client` instead of the `requests` library which is already a dependency (see: https://github.com/ooni/api/blob/cbf7a3925812d01af798e70cb3f71de153dffe2f/measurements/api/measurements.py#L184)?

FedericoCeratto: https://docs.python.org/3.7/library/urllib.request.html and http.client provide most of the same features without needing an external dependency. We could slowly move to urllib.request over time and drop the dep.

measurements/api/measurements.py
FedericoCeratto: Done

measurements/api/measurements.py
hellais: Ok seems reasonable. Let's keep this as-is then for the time being and maybe consider dropping requests in the future.

See: https://github.com/ooni/api/issues/130

hellais:
Merged! 🎉

PR #139, opened 2019-12-05
Add throttling

FedericoCeratto:
Related to #116 and #117

### DIFF
--- a/measurements/app.py
+++ b/measurements/app.py
@@ -8,16 +8,15 @@ import datetime
 import sys
 import os
 
-from flask import Flask, json
+from flask import Flask, json, request
 from flask_misaka import Misaka
 from flask_cors import CORS
+import flask_limiter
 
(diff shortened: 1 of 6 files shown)


measurements/app.py
```diff
@@ -94,6 +93,16 @@ def init_app(app, testmode=False):
 def check_config(config):
     pass
 
+def extract_client_ipaddr_for_throttling():
+    ipaddr = flask_limiter.util.get_ipaddr()
+    # TODO: move to an external config file /etc/ooniapi.conf ?
+    explorer_addr = "37.218.245.0"
```
hellais: Beware that the last octet is getting sanitized by our nginx config, so the IP of the explorer likely has an extra octet in here.

FedericoCeratto: The last .0 octet is due to that.

PR #140, opened 2019-12-06
Fix #355

FedericoCeratto:
Add filtering on confirmed and anomaly in list_measurements for fastpath.
This is meant to be temporary. I'll add anomaly and confirmed columns to the fastpath table and populate them to ease integration with the API, Explorer and manually run queries - for the time being.

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -539,6 +539,9 @@ def list_measurements(
         fpq = fpq.filter(Fastpath.test_name == test_name)
     if input_:
         fpq = fpq.filter(Fastpath.input == input_)
+    # TODO, we don't support filtering by confirmed filter in the fastpath, so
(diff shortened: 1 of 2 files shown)


measurements/api/measurements.py
```diff
@@ -537,7 +539,12 @@ def list_measurements(
         fpq = fpq.filter(Fastpath.test_name == test_name)
     if input_:
         fpq = fpq.filter(Fastpath.input == input_)
-
+    if confirmed is not None:
+        fpq = fpq.filter(False)
```
hellais: This means basically that we can't filter by confirmed on the fastpath, correct?

Maybe we should add a comment above this line stating that is what is going on here.

FedericoCeratto: Correct. I put a comment in the PR - the idea is to then add the confirmed column.

PR #141, opened 2019-12-09
Add support for fastpath anomaly/confirmed/failure cols

FedericoCeratto:
To be deployed only after https://github.com/ooni/pipeline/pull/271

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -469,7 +469,12 @@ def list_measurements(
         mrwhere.append(sql.text("report.test_name = :test_name"))
         fpwhere.append(sql.text("test_name = :test_name"))
 
-    # anomaly and confirmed: any value != TRUE is treated as FALSE
+    # Filter on anomaly, confirmed and failure:
+    # The database stores anomaly and confirmed as boolean + NULL and stores
+    # failures in different columns. This leads to many possible combinations
+    # but only a subset is used.
+    # On anomaly and confirmed: any value != TRUE is treated as FALSE
+    # See test_list_measurements_filter_flags_fastpath
 
     if anomaly is True:
         mrwhere.append(sql.text("measurement.anomaly IS TRUE"))
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -17,8 +17,7 @@ from werkzeug.exceptions import HTTPException, BadRequest
 
 from sqlalchemy import func, and_, false, text, select, sql, column
 from sqlalchemy.sql import literal_column
-from sqlalchemy import String, cast, Float
-from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
+from sqlalchemy import String, cast
 from sqlalchemy.exc import OperationalError
 from psycopg2.extensions import QueryCanceledError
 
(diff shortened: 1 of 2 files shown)


hellais:
This may have to be dropped in light of the new changes to the data model

hellais:
I am removing myself as a reviewer and I suspect we will have to close this PR. @FedericoCeratto I will leave that decision to you.

FedericoCeratto:
The changes in fastpath in https://github.com/ooni/pipeline/pull/271 are deployed. Rebasing and updating this PR.

FedericoCeratto:
@hellais the PR is now up to date (essentially rewritten from scratch)

measurements/api/measurements.py
```diff
@@ -482,21 +469,34 @@ def list_measurements(
...
-        fpwhere.append(sql.text("false"))
+    # anomaly and confirmed: any value != TRUE is treated as FALSE
+
+    if anomaly is True:
+        mrwhere.append(sql.text("measurement.anomaly IS TRUE"))
+        fpwhere.append(sql.text("fastpath.anomaly IS TRUE"))
+
+    elif anomaly is False:
+        mrwhere.append(sql.text("measurement.anomaly IS NOT TRUE"))
+        fpwhere.append(sql.text("fastpath.anomaly IS NOT TRUE"))
```
hellais: Given the `anomaly` is nullable, saying `anomaly = False` is not the same as `anomaly != True`.

I suggest changing this to:
```
mrwhere.append(sql.text("measurement.anomaly IS FALSE"))
```

measurements/api/measurements.py
```diff
@@ -482,21 +469,34 @@ def list_measurements(
...
+
+    if confirmed is True:
+        mrwhere.append(sql.text("measurement.confirmed IS TRUE"))
+        fpwhere.append(sql.text("fastpath.confirmed IS TRUE"))
+
+    elif confirmed is False:
+        mrwhere.append(sql.text("measurement.confirmed IS NOT TRUE"))
+        fpwhere.append(sql.text("fastpath.confirmed IS NOT TRUE"))
```
hellais: Same as above

measurements/api/measurements.py
```diff
@@ -482,21 +469,34 @@ def list_measurements(
...
+
+    if failure is True:
+        # residual_no is never NULL, msm_failure is always NULL
+        mrwhere.append(sql.text("measurement.exc IS NOT NULL"))
+        fpwhere.append(sql.text("fastpath.msm_failure IS TRUE"))
+
+    elif failure is False:
+        # on success measurement.exc is NULL
+        mrwhere.append(sql.text("measurement.exc IS NULL"))
+        fpwhere.append(sql.text("fastpath.msm_failure IS NOT TRUE"))
```
hellais: Same as above

measurements/api/measurements.py
FedericoCeratto: We are treating NULL as an absence of anomaly (hence we coalesce FALSE), so if someone is querying the API using anomaly=False we should return any row that is not TRUE

FedericoCeratto:
rebased

measurements/api/measurements.py
hellais: Ok fair enough. We are pretty light on details in the spec, but I am pretty sure this is a behaviour change compared to the previous API version.

I am fine doing this, but I figured I would flag it to be sure we this is what we want.

PR #144, opened 2019-12-10
Add domain filter to fastpath. Close #142

### LINKED ISSUE #142: list_measurements is not filtering on domain on fastpath table

FedericoCeratto:


measurements/api/measurements.py
```diff
@@ -539,6 +539,18 @@ def list_measurements(
...
+        if test_name in [None, "web_connectivity", "http_requests"]:
+            fpq = fpq.filter(
+                or_(
+                    Fastpath.input.startswith(domain),
+                    Fastpath.input.startswith("http://" + domain),
+                    Fastpath.input.startswith("https://" + domain),
+                )
+            )
+        else:
+            fpq = fpq.filter(Fastpath.input.startswith(domain))
```
FedericoCeratto: 5Looks like this code path is not used by Explorer: domain is set while test_name is not in [None, "web_connectivity", "http_requests"]

Should it run or return error?

hellais: It should run (I think), because you want to be be able to filter for results of tests like the `dns_consistency` or `tcp_connect` tests which have as a target just a bare hostname

Issue #145, opened 2019-12-16
Avoid polluting error log when we skip the rate limiting


hellais:
I logged into sentry today and found 237k events triggered by:
```
Skipping limit: 400 per 1 hour. Empty value found in parameters.
```

and similar stuff.

Many of these errors were triggered by the rate limiting acting on a request which should be limited, but several were being triggered by our server-side loads which should be whitelisted (see: https://github.com/ooni/api/blob/master/measurements/app.py#L99).

Digging a bit more into the issue I was able to confirm that in fact it's not actually applying a limit to these requests, but it's just that `flask_limiter` treats a `None` return value from the limit function as an error: https://github.com/alisaifee/flask-limiter/blob/master/flask_limiter/extension.py#L390.

Apparently this issue was reported a year ago on the flask_limiter repo to no avail: https://github.com/alisaifee/flask-limiter/issues/186.

I suggest as a quick solution to find a way to disable the error logging of flask_limiter so that we stop at least spamming sentry, but possibly also upstream a patch with the correct fix.

PR #146, opened 2019-12-16
Disable recording error logs from flask-limiter

hellais:
Fixes: https://github.com/ooni/api/issues/145


hellais:
Added a note about that in the comment

PR #150, opened 2019-12-26
Refactor list_measurements

FedericoCeratto:
- Move away from SQLAlchemy models
- Fix logic to merge measurements between pipeline and fastpath
- Many big speed improvements
- Order by measurement_start_time by default
- Add support for domain_input
- Add commented-out, but tested support for category_code
- Increase test coverage

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -17,6 +17,7 @@ from werkzeug.exceptions import HTTPException, BadRequest
 
 from sqlalchemy.dialects import postgresql
 from sqlalchemy import func, or_, and_, false, text, select, sql, column
+from sqlalchemy.sql import literal_column
 from sqlalchemy import String, cast, Float
 from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 from sqlalchemy.exc import OperationalError
@@ -28,8 +29,6 @@ from measurements import __version__
 from measurements.config import REPORT_INDEX_OFFSET, REQID_HDR, request_id
 from measurements.models import (
     Autoclaved,
-    Domain_input,
-    Fastpath,
     Input,
     Measurement,
     Report,
@@ -268,39 +267,12 @@ def get_measurement(measurement_id, download=None):
     return response
 
 
-def input_cte(input_, domain, test_name):
-    """Given a domain or an input_, build a WHERE filter
-    """
-    if input_ and domain:
-        raise BadRequest("Must pick either domain or input")
-
-    if not input_ and not domain:
-        return None
-
-    if input_:
-        # FIXME: bug on left "%"
-        where_or = [text("input.input LIKE :i").bindparams(i="%{}%".format(input_))]
-        url_q = (
-            select(
-                [column("input").label("input"), column("input_no").label("input_no")]
-            )
-            .where(or_(*where_or))
-            .select_from(sql.table("input"))
-        )
-        return url_q.cte("input_cte")
-
-    # Filter by domain using the domain_input table
-    url_q = (
-        select([column("input").label("input"), column("input_no").label("input_no")])
-        .where(column("domain") == domain)
-        .select_from(sql.table("domain_input"))
-    )
-
-    return url_q.cte("input_cte")
-
-
 def log_query(log, q):
     import sqlparse  # debdeps: python3-sqlparse
+
+    if "Select" in repr(type(q)): # UGLY
+        log.info("\n--- query ---\n\n%s\n\n-------------", q)
+        return
 
     sql = str(
         q.statement.compile(
@@ -364,6 +336,7 @@ def list_measurements(
     failure=None,
     anomaly=None,
     confirmed=None,
+    category_code=None,
 ):
     """Search for measurements using only the database. Provide pagination.
     """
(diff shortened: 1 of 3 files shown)


measurements/api/measurements.py
```diff
@@ -261,50 +267,13 @@ def get_measurement(measurement_id, download=None):
...
-        .select_from(sql.table("input"))
-    )
-
-    return url_q.cte("input_cte")
-
-
 def log_query(log, q):
     import sqlparse  # debdeps: python3-sqlparse
 
+    if "Select" in repr(type(q)):  # UGLY
```
hellais: Why is this being done?

Maybe we can add a comment here to explain it.

Perhaps it's possible to rather do something like `isinstance()` or `issubclass`

measurements/api/measurements.py
```diff
@@ -420,171 +390,226 @@ def list_measurements(
...
+
+    if failure is True:
+        mrwhere.append(sql.text("measurement.exc IS NOT NULL"))
+        mrwhere.append(sql.text("measurement.residual_no IS NOT NULL"))
+        mrwhere.append(sql.text("measurement.msm_failure IS TRUE"))
+
+    if failure is False:
+        mrwhere.append(sql.text("measurement.exc is NULL"))
+        mrwhere.append(sql.text("measurement.residual_no is NULL"))
+        mrwhere.append(sql.text("measurement.msm_failure is FALSE"))
```
hellais: These two are not equivalent to how they were before, because they are now getting considered as an `AND` chain, while previously these were an `OR`.

I suggest you either do:
```
mrwhere.append(sql.text("measurement.exc IS NOT NULL OR measurement.residual_no IS NOT NULL OR measurement.msm_failure IS TRUE"))
```
thought I am not sure this works

Or you do
```
or_([
        sql.text("measurement.exc IS NOT NULL"),
        sql.text("measurement.residual_no IS NOT NULL"),
        sql.text("measurement.msm_failure IS TRUE")
])
```

measurements/api/measurements.py
```diff
@@ -420,171 +390,226 @@ def list_measurements(
...
+        fpwhere.append(sql.text("false"))
+
     if anomaly is not None:
-        fpq = fpq.filter(
-            cast(cast(Fastpath.scores["blocking_general"], String), Float) > 0.5
+        # TODO: coalesce false() ?
+        query_params["anomaly"] = anomaly
+        mrwhere.append(sql.text("measurement.anomaly = :anomaly"))
+        # fpwhere.append(sql.text("scores->>'blocking_general' > 0.5"))
+        fpwhere.append(sql.text("false"))
```
hellais: I think we should add a comment here saying that currently we exclude from the filter the fastpath table as it doesn't have a confirmed or anomaly column and perhaps reference the relevant issue.

measurements/api/measurements.py
```diff
@@ -420,171 +390,226 @@ def list_measurements(
...
 
+    else:
+        mr_table = mr_table.join(
+            sql.table("domain_input"),
+            sql.text("domain_input.input_no = measurement.input_no"),
+        )
+
+    # We runs SELECTs on the measurement-report (mr) tables and faspath independently
+    # from each other and then merge them.
+    # An ORDER BY + LIMIT on "limit+offset" is applied in each SELECT as a speedup.
```
hellais: I wonder if an additional few words are necessary to clarify the significance of using offset + limit, as it was not obvious to me in the beginning we should do that.

Something to say that we need to retrieve all the records up to limit + offset and then apply the offset and limit on the outer query.

FedericoCeratto:
(rebased on master)

measurements/api/measurements.py
FedericoCeratto: Removed entirely :)

measurements/api/measurements.py
FedericoCeratto: Thanks! Fixed in 77751d1a1f418ff6e4081a42f092a8f58f3321a9 and added tests.

measurements/api/measurements.py
FedericoCeratto: It's going to be fixed very soon in a new PR. I'll reference this one.

measurements/api/measurements.py
FedericoCeratto: Added comment in ff7529b188a622bdcde80aed2f56ec7f39d9dbc0

measurements/api/measurements.py
```diff
@@ -420,171 +374,228 @@ def list_measurements(
...
 
-    fpq = current_app.db_session.query(*fpcols)
+    mrwhere = []
+    fpwhere = []
+    query_params = {}
+
+    # Populate WHERE clauses and query_params dict
+
+    if failure is True:
+        # residual_no is never NULL, msm_failure is always NULL
```
hellais: Hum. The fact that `msm_failure` is always `NULL` sounds like a bug.

I believe that `exc` is used to track exceptions in centrifugation of the measurement file, so as the logic for processing measurements gets better, this should in theory be cleaned up.

In light of the fact that the `failure` value is being set to the `msm_failure` column (https://github.com/ooni/api/blob/master/measurements/api/measurements.py#L426), maybe we should keep the filter as being applied to the msm_failure and document the fact that it's always NULL as a bug in the pipeline?

FedericoCeratto: @hellais my suggestion is to ignore residual_no and msm_failure for the moment so that the API queries run faster and the code is leaner. If/when we fix the colums we can add the check back in the API.

hellais: Ok fair enough.

measurements/api/measurements.py
hellais: I filed this issue: https://github.com/ooni/api/issues/157

PR #153, opened 2020-01-02
Remove SQLAlchemy models

FedericoCeratto:
Remove SQLalchemy models from list_files
Add integration tests
Speed improvements: count(*) should be faster now
Bump up requirements


FedericoCeratto:
(still WIP)

measurements/api/measurements.py
```diff
@@ -206,41 +221,43 @@ def get_measurement(measurement_id, download=None):
...
-    except NoResultFound:
-        # XXX we should actually return a 404 here
-        raise BadRequest("No measurement found")
+    )
+    where = sql.text("measurement.msm_no = :msm_no")
+    query = select(cols).where(where).select_from(table)
+    query_params = dict(msm_no=msm_no)
+    q = current_app.db_session.execute(query, query_params)
+
+    msmt = q.fetchone()
```
hellais: Should be noted we aren't handling the case in which there are duplicate results. We may or may not care about that.

In theory it's easy to fix by just doing an additional call to `fetchone()` and ensuring we get `None` on the second call.

measurements/api/private.py
```diff
@@ -165,169 +161,42 @@ def api_private_countries():
...
-                "probe_asn": "AS{}".format(row.probe_asn),
-                "test_start_time": row.test_start_time,
-                "input": row.input,
-            }
-        )
 
-    return jsonify({"results": results})
+@api_private_blueprint.route("/blockpages", methods=["GET"])
+def api_private_blockpages():
+    abort(404)
```
hellais: Maybe we also want to log this in a structured way? Though OTOH perhaps 404 is enough as we can easily parse nginx logs looking for that.

measurements/api/measurements.py
FedericoCeratto: Until we introduce OOID we cannot pinpoint msmts and get_measurement is simply returning the row.

measurements/api/private.py
FedericoCeratto: We are not logging other calls e.g. /test_names (except from the Nginx side). Also the deprecated calls seemed to have no traffic at all.

FedericoCeratto:
@hellais updated after removing most of models.py and rerunning tests

PR #160, opened 2020-01-15
Implement smarter order_by in list_measurements

FedericoCeratto:
Add benchmark tests

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -330,7 +330,7 @@ def list_measurements(
     since=None,
     until=None,
     since_index=None,
-    order_by="measurement_start_time",
+    order_by=None,
     order="desc",
     offset=0,
     limit=100,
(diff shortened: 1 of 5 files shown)


measurements/api/measurements.py
```diff
@@ -569,8 +569,23 @@ def list_measurements(
         .select_from(fpq_table)
         .limit(offset + limit)
     )
-    assert order_by  # TODO: is this always true? Was it a bug?
-    if order_by is not None:
+
+    if order_by is None:
```
hellais: Let's also be sure to update: https://github.com/ooni/api/blob/master/measurements/openapi/measurements.yml#L111 to reflect these changes.

I think we should probably deprecate setting the order_by field entirely and always enforce ordering by measurement_start_time or test_start_time.

Perhaps to ensure clients notice, we can say that the only allowed value for `order_by` is `deprecated`.

FedericoCeratto: Updated. The deprecation has been discussed on Slack and can be done in a different PR.

PR #162, opened 2020-02-10
Generate metrics on query runtime

FedericoCeratto:
ooni/backend#325
QueryCanceledError were handled as API errors leading to a traceback
being uploaded to Sentry. A query timeout is not an API bug or
error condition.
We now capture all query runtimes in a metric for prometheus.
The query statement is too long to be used as query name or label,
so we log a short hash instead.

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -9,7 +9,7 @@ from dateutil.parser import parse as parse_date
 import requests
 import lz4framed
 
-from sentry_sdk import configure_scope, capture_exception
+import sentry_sdk as sentry
 
 from flask import current_app, request, make_response, abort
 from flask.json import jsonify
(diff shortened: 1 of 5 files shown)


measurements/api/measurements.py
```diff
@@ -628,69 +628,73 @@ def coal(colname):
...
+                    "confirmed": row.confirmed,
+                    "failure": row.failure,
+                    "scores": json.loads(row.scores),
+                }
             )
+    except OperationalError as exc:
+        if isinstance(exc.orig, QueryCanceledError):
+            # Timeout due to a slow query. Generate metric and do not feed it
+            # to Sentry.
+            raise QueryTimeoutError()
```
hellais: By raising the exceptions aren't you sending it to sentry?

measurements/database.py
```diff
@@ -16,6 +17,14 @@
 
 from measurements.config import request_id
 
+from flask import Flask
+from prometheus_flask_exporter import PrometheusMetrics
+from prometheus_client import Gauge
+
+app = Flask(__name__)
+registry = PrometheusMetrics(app).registry
+query_time = Gauge("query", "query", ["hash",], registry=registry)
```
hellais: I think this code should be moved into the app initialisation. I don't think you should be duplicating the Flask app instance, as the initialisation is meant to be done only once.

measurements/api/measurements.py
FedericoCeratto: That was my question on Slack: is FlaskIntegration in app.py acting as a catch-all? If so, the previous implementation of timeout handling was unnecessary.

measurements/api/measurements.py
```diff
@@ -628,69 +628,73 @@ def coal(colname):
     query = select(merger).select_from(j).order_by(fob).offset(offset).limit(limit)
     log.info(query)
 
+    with sentry.configure_scope() as scope:
+        # Set query (without params) in Sentry scope for the rest of the API call
+        # https://github.com/getsentry/sentry-python/issues/184
+        scope.set_extra("sql_query", query)
```
hellais: Oh lol, this is interesting.

PR #165, opened 2020-02-14
Add query execution limits

FedericoCeratto:
Prevent hanging queries. Better to time out on the database than in the browser.

### DIFF
--- a/measurements/database.py
+++ b/measurements/database.py
@@ -48,6 +48,10 @@ def init_db(app):
     Base.query = app.db_session.query_property()
     init_query_logging(app)
 
+    # Set query duration limits (in milliseconds)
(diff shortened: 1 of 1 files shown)


FedericoCeratto:
Prevent hanging queries. Better to time out on the database than in the browser.

measurements/database.py
```diff
@@ -48,6 +48,10 @@ def init_db(app):
     Base.query = app.db_session.query_property()
     init_query_logging(app)
 
+    # Set query duration limits (in milliseconds)
+    app.db_session.execute("SET statement_timeout = 30000")
```
hellais: Based on what was 30 seconds chosen?

FedericoCeratto: ...TCP socket timeouts in browsers

PR #166, opened 2020-02-19
get_measurement_meta: initial implementation

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/341

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -144,7 +144,8 @@ def list_files(
                 "test_start_time": row.test_start_time,
             }
         )
-    # We got less results than what we expected, we know the count and that we are done
+    # We got less results than what we expected, we know the count and that we
+    # are done
     if len(results) < limit:
         count = offset + len(results)
         pages = math.ceil(count / limit)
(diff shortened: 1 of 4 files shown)


measurements/api/measurements.py
```diff
@@ -321,6 +321,118 @@ def _merge_results(tmpresults):
     return tuple(resultsmap.values())
 
 
+def get_measurement_meta(report_id=None, input=None):
+    """Get metadata on one measurement by measurement_id + input
+    """
```
hellais: You probably want to ensure that `report_id` is not NULL, either in here or in the swagger spec.

measurements/api/measurements.py
```diff
@@ -321,6 +321,118 @@ def _merge_results(tmpresults):
...
+    )
+    query = select(merger).select_from(j)
+
+    with sentry.configure_scope() as scope:
+        # Set query (without params) in Sentry scope for the rest of the API call
+        # https://github.com/getsentry/sentry-python/issues/184
+        scope.set_extra("sql_query", query)
+
+    q = current_app.db_session.execute(query, query_params)
+    x = q.fetchone()
```
hellais: What happens if more than one row is returned? Do we want to somehow handle that case and log it somehow?

measurements/api/measurements.py
```diff
@@ -321,6 +321,118 @@ def _merge_results(tmpresults):
...
+    return dict(x)
```
hellais: It may improve readability if the exact keys of the dict are expanded to make it explicit.

tests/integ/test_integration.py
```diff
@@ -302,6 +302,63 @@ def test_list_files_range_cc_asn(client):
     assert len(results) == 7
 
 
+# # get_measurement_meta # #
+
```
hellais: I would also add a test case for the scenario of a measurement which doesn't have an input.

measurements/api/measurements.py
FedericoCeratto: Done

measurements/api/measurements.py
FedericoCeratto: Added comment

measurements/api/measurements.py
FedericoCeratto: There are 24 columns listed in "merger" few lines above. Repeating them would be brittle as we add/remove cols in future.

tests/integ/test_integration.py
FedericoCeratto: Yes - added support in the last commit.

FedericoCeratto:
Superseded

PR #167, opened 2020-02-21
Allow keeping rows in faspath table

FedericoCeratto:
In list_measurement a msmt both in mr_table and fastpath gets its
measurement_id from mr_table and not from fastpath in order to use
data from S3 and allow deleting JSON files on the fastpath host.

Dependency of https://github.com/ooni/backend/issues/233

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -605,10 +605,13 @@ def list_measurements(
             literal_column(f"fp.{colname}"), literal_column(f"mr.{colname}")
         ).label(colname)
 
+    # Merge data from mr_table and fastpath.
+    # Most of the time we prefer data from fastpath, using coal().
+    # For measurement_id, we prefer mr_table. See test_list_measurements_shared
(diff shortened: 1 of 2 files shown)


measurements/api/measurements.py
```diff
@@ -605,10 +605,13 @@ def coal(colname):
...
         ).label(colname)
 
+    # Merge data from mr_table and fastpath.
+    # Most of the time we prefer data from fastpath, using coal().
+    # For measurement_id, we prefer mr_table. See test_list_measurements_shared
     merger = [
         coal("test_start_time"),
         coal("measurement_start_time"),
-        coal("measurement_id"),
+        func.coalesce(literal_column("mr.measurement_id"), literal_column("fp.measurement_id")).label("measurement_id"),
```
hellais: This makes sense. Maybe it's worth mentioning in the docstring that `mr.measurement_id` and `fp.measurement_id` have a different prefix.

PR #171, opened 2020-02-25
List measurements anomaly coalesce

FedericoCeratto:
Add list_measurements_anomaly_coalesce test
Improve list_measurements outer join merging
Simplify _merge_results: remove _merge_two_results
Add integration tests

NOTE: to be rebased on master

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -626,7 +626,6 @@ def list_measurements(
     # offset
     fob = text("{} {}".format(order_by, order))
     query = select(merger).select_from(j).order_by(fob).offset(offset).limit(limit)
-    log.info(query)
 
     with sentry.configure_scope() as scope:
         # Set query (without params) in Sentry scope for the rest of the API call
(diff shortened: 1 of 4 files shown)


measurements/api/measurements.py
```diff
@@ -668,10 +643,16 @@ def coal(colname):
...
-    # traditional pipeline and one from fastpath, merged together
+    # For each report_id / input tuple, we want at most one entry. Measurements
+    # from mr_table and fastpath has already been merged by the FULL OUTER JOIN
+    # but we have duplicate msmts sharing the same report_id / input.
     results = _merge_results(tmpresults)
 
+    # Replace the special value INULL for "input" with None
+    for i, r in enumerate(results):
+        if r["input"] == INULL:
+            results[i]["input"] = None
```
hellais: Is this done because otherwise the explorer links will not work as the input is set to ""?

Maybe it's useful to also mention that INULL is equal to the empty string in the docstring.

FedericoCeratto: https://github.com/ooni/api/pull/171/files/7489bda785d4d3a193a8252262b5be0fb0f75a1e#diff-a6bf0b1b75ef5e2f06d17e0a4e9bbc4aR367  it's described here. Currently the empty string seems to work safely but we could use a different magic value if needed.

PR #177, opened 2020-03-12
Drop the use of count() for list_files endpoint

hellais:
This implements: https://github.com/ooni/backend/issues/356


FedericoCeratto:
Partially related to https://github.com/ooni/sysadmin/issues/433

PR #178, opened 2020-03-17
Aggegation API

FedericoCeratto:
Depends on https://github.com/ooni/pipeline/pull/304
Implements https://github.com/ooni/backend/issues/321

### DIFF
--- a/measurements/api/measurements.py
+++ b/measurements/api/measurements.py
@@ -4,7 +4,9 @@ import math
 import re
 import time
 
+from csv import DictWriter
 from dateutil.parser import parse as parse_date
+from io import StringIO
 
 import requests
 import lz4framed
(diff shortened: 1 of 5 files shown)


measurements/openapi/measurements.yml
```diff
@@ -249,6 +249,161 @@ paths:
...
+            {"cnt": 25, "measurement_start_day": "2020-01-27"},
+        ],
+        "v": 0,
+    }
+    assert r == exp, fjd(r)
+
+
+def test_aggregation_x_axis_y_axis(client, log):
+    # 2 dimensions
+    url = "aggregation?since=2020-01-01&until=2020-02-01&axis_x=measurement_start_day&axis_y=probe_cc&test_name=web_connectivity"
```
hellais, line 404: Is this a typo? I don't think python code should be inside of the swagger spec.

measurements/openapi/measurements.yml
```diff
@@ -249,6 +249,161 @@ paths:
...
+          - multi_protocol_traceroute
+          - meek_fronted_requests_test
+          - whatsapp
+          - vanilla_tor
+          - facebook_messenger
+          - ndt
+          - dash
+          - telegram
+          - psiphon
+          - tor
```
hellais, line 302: Maybe we want to define this as a special type so that we don't have to update the test_name list everywhere each time we add a new one.

measurements/api/measurements.py
```diff
@@ -712,3 +714,139 @@ def coal(colname):
...
+
+        else:
+            r = dict(q.fetchone())
+
+        if format == "CSV":
+            return _convert_to_csv(r)
+
+        response = jsonify({"v": 0, "dimension_count": dimension_cnt, "result": r})
+        if cacheable:
+            response.cache_control.max_age = 3600 * 24
```
hellais, line 848: This is good.

PR #179, opened 2020-03-20
Implement custom rate limiter

FedericoCeratto:
Supports daily, weekly and monthly quotas. Quotas are measured in API usage time.
Provide stubs to add token-based quotas.

Related to https://github.com/ooni/backend/issues/371

### DIFF
--- a/rate_limit_quotas.py
+++ b/rate_limit_quotas.py
@@ -85,9 +85,6 @@ class Limiter:
                 b = self._ipaddr_buckets[n]
                 b[ipaddr] = b.get(ipaddr, limit) - elapsed
 
-        print([b[ipaddr] for b in self._ipaddr_buckets])
-        self.increment_quota_counters(1)
-
     def get_minimum_across_quotas(self, ipaddr=None, token=None) -> int:
         if ipaddr:
             iterable = zip(self._ipaddr_limits, self._ipaddr_buckets)
--- a/rate_limit_quotas.py
+++ b/rate_limit_quotas.py
@@ -1,0 +1,186 @@
+"""
+Rate limiter and quota system.
+
+Framework-independent rate limiting mechanism that provides:
+    * IP address and token-based accounting
+    * customizable quotas based on IP address and token
+    * late limiting based on resource usage (time spent on API calls)
+    * bucketing based on day, week, month
+    * statistics
+    * metrics
+    * fast in-memory storage
+
+Also provides a connector for Flask
+
+"""
+
+import time
+import ipaddress
+from typing import Dict
+
+
+class Limiter:
+    def __init__(
+        self,
+        limits: dict,
+        token_check_callback=None,
+        ipaddr_methods=["X-Real-Ip", "socket"],
+    ):
+        # Bucket sequence: month, week, day
+        self._ipaddr_limits = [
+            limits.get(l, None)
+            for l in ("ipaddr_per_month", "ipaddr_per_week", "ipaddr_per_day")
+        ]
+        self._token_limits = [
+            limits.get(l, None)
+            for l in ("token_per_month", "token_per_week", "token_per_day")
+        ]
+        self._ipaddr_buckets = ({}, {}, {})  # type: Dict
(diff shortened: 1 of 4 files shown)


rate_limit_quotas.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def consume_quota(self, elapsed: float, ipaddr=None, token=None) -> None:
+        """Consume quota in seconds
+        """
+        if ipaddr:
+            for n, limit in enumerate(self._ipaddr_limits):
+                b = self._ipaddr_buckets[n]
+                b[ipaddr] = b.get(ipaddr, limit) - elapsed
+
+        print([b[ipaddr] for b in self._ipaddr_buckets])
```
hellais: I suppose this is left over from debugging. Maybe we can either remove it or use the proper logging module for it?

rate_limit_quotas.py
```diff
@@ -0,0 +1,186 @@
...
+
+from flask import request, current_app
+import flask
+
+
+class FlaskLimiter:
+    def _get_client_ipaddr(self):
+        # https://github.com/alisaifee/flask-limiter/issues/41
+        for m in self._limiter._ipaddr_extraction_methods:
+            if m == "X-Forwarded-For":
```
hellais: In `ipaddr_methods` it's mentioned only `["X-Real-Ip", "socket"]`, but `X-Forwarded-For` is not present, is that intended?

rate_limit_quotas.py
```diff
@@ -0,0 +1,186 @@
...
+        if ipaddr:
+            for n, limit in enumerate(self._ipaddr_limits):
+                b = self._ipaddr_buckets[n]
+                b[ipaddr] = b.get(ipaddr, limit) - elapsed
+
+        print([b[ipaddr] for b in self._ipaddr_buckets])
+        self.increment_quota_counters(1)
+
+    def get_minimum_across_quotas(self, ipaddr=None, token=None) -> int:
+        if ipaddr:
```
hellais: Should we maybe handle the case in which `ipaddr` is not set? Is there any way this function could get called without a valid `ipaddr` value?

rate_limit_quotas.py
```diff
@@ -0,0 +1,186 @@
...
+    def _after_request_callback(self, response):
+        """Consume quota and injects HTTP headers when responding to a request
+        """
+        log = current_app.logger
+        try:
+            assert response
+            tdelta = time.monotonic() - self._request_start_time
+            ipaddr = self._get_client_ipaddr()
+            self._limiter.consume_quota(tdelta, ipaddr=ipaddr)
+            q = self._limiter.get_minimum_across_quotas(ipaddr=ipaddr)
```
hellais: Maybe the missing ipaddr value can be handled here too.

rate_limit_quotas.py
```diff
@@ -0,0 +1,186 @@
...
+                v += vdelta
+                if v >= limit:
+                    to_delete.append(k)
+                else:
+                    bucket[k] = v
+
+            for k in to_delete:
+                del bucket[k]
+
+    def refresh_quota_counters_if_needed(self):
```
hellais: I see that this is only called in the constructor, shouldn't it be called also periodically in some way?

rate_limit_quotas.py
FedericoCeratto: Yes, fixed.

rate_limit_quotas.py
FedericoCeratto: Yes: we are currently using X-Real-Ip set by Nginx

rate_limit_quotas.py
FedericoCeratto: Not a risk with the current codebase, but I'm adding assertions around this.

rate_limit_quotas.py
FedericoCeratto: Yes, this will be added externally to the module.

hellais:
I see this branch is still in draft state. I also don't see any new commits since my comments, @FedericoCeratto did you maybe forget to push?

PR #188, opened 2020-05-04
Rate limiting: add quotas whitelisting and summary

FedericoCeratto:
General cleanup and added typing

### DIFF
--- a/measurements/app.py
+++ b/measurements/app.py
@@ -100,23 +100,6 @@ def check_config(config):
     pass
 
 
-def extract_client_ipaddr_for_throttling():
-    # TODO: move addrs to an external config file /etc/ooniapi.conf ?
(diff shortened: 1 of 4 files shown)


measurements/app.py
```diff
@@ -144,7 +127,10 @@ def create_app(*args, testmode=False, **kw):
         ipaddr_per_day=400,
         token_per_day=500,
     )
-    app.limiter = FlaskLimiter(limits=limits, app=app,)
+    # Whitelist Prometheus and AMS Explorer
+    # TODO: move addrs to an external config file /etc/ooniapi.conf ?
+    whitelist = ["37.218.245.43", "37.218.242.149"]
```
hellais, line 132: Maybe we can use a domain name and lookup the IPs at start time?

PR #190, opened 2020-05-27
Aggregation: return axis names, filter columns

FedericoCeratto:
Minor code cleanup

### DIFF
--- a/measurements/openapi/measurements.yml
+++ b/measurements/openapi/measurements.yml
@@ -353,12 +353,26 @@ paths:
           type: string
           description: |
             The dimension on the x axis e.g. measurement_start_day
+          enum:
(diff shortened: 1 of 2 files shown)


measurements/openapi/measurements.yml
```diff
@@ -353,12 +353,26 @@ paths:
...
           type: string
           description: |
             The dimension on the y axis e.g. probe_cc
+          enum:
+          - category_code
+          - input
+          - measurement_start_day
+          - probe_asn
+          - probe_cc
+          - test_name
```
hellais, line 375: I think it would be helpful to have in here also a sample output document

FedericoCeratto:
To be reviewed and rebased

PR #193, opened 2020-08-25
Update css file name from ooni.org

sarathms:
Fixes ooni/backend#440

@hellais For the missing fonts problem, the best solution is to configure CORS policy on `ooni.org`


sarathms:
Also, this dependency on the CSS filename is not a good idea. There should be a more auto

bassosimone (approved):
🐳

PR #196, opened 2020-11-03
Upgrade macos

hellais:
Tested to work with macOS 10.15.7


hellais:
This is superseeded by https://github.com/ooni/api/pull/220

PR #207, opened 2020-12-14
0.15 Bump up test-list/urls default limit

FedericoCeratto:
https://github.com/ooni/probe/issues/1311


newapi/ooniapi/prio.py
```diff
@@ -207,7 +207,7 @@ def list_test_urls():
         category_codes = tuple(category_codes)
         limit = int(param("limit") or -1)
         if limit == -1:
-            limit = 100
+            limit = 3000
```
hellais, line 210: Can we instead apply no limit?

FedericoCeratto: Why?

PR #213, opened 2021-02-08
0.20 Support legacy tests in _fetch_measurement_body

FedericoCeratto:


newapi/ooniapi/measurements.py
```diff
@@ -289,13 +291,18 @@ def _fetch_autoclaved_measurement_body(report_id: str, input) -> dict:
...
+    if u_count == 5:
         # Look on disk and then from JSONL cans on S3
         body = _fetch_measurement_body_on_disk(report_id, input)
         if body is None:
             log.debug(f"Fetching body for {report_id} {input} from jsonl on S3")
             body = _fetch_jsonl_measurement_body(report_id, input)
-    elif report_id.count("_") == 2:
+    elif u_count == 2:
+        body = _fetch_autoclaved_measurement_body(report_id, input)
+    elif u_count == 1:
```
hellais, line 303: Are we sure the underscores are a good indicator of which client generated the test? If so we should probably document in here what the mappings of underscore count to probe_version looks like.

hellais (approved):
Left a minor nitpick comment, otherwise LGTM

newapi/ooniapi/measurements.py
FedericoCeratto: Added comment

PR #217, opened 2021-02-19
Fix top image in new api

KwadroNaut:
Broken image in the current deployed version.


sarathms (approved):
LGTM! Thanks @KwadroNaut

PR #218, opened 2021-02-25
Sync client and server API definition

bassosimone:
With the changes in this diff, we have no differences between the client and the server's view of the API.

This branch uses https://github.com/ooni/api/pull/219, which really helped me to run the API locally and fix all the issues.

See https://github.com/ooni/probe/issues/1355


FedericoCeratto (approved):
Discussed on slack, LGTM

bassosimone:
Thank you!

PR #220, opened 2021-03-01
Fixup to work under macOS

hellais:


README.md
```diff
@@ -48,6 +48,20 @@ cd newapi
 ./spawnrunner pytest-3 tests/integ/test_probe_services.py
 ```
 
+### Running the API using gunicorn on macOS
+
+First setup a local port forward with:
+
+```bash
+ssh ams-pg-test.ooni.org -L 0.0.0.0:15432:127.0.0.1:5432 -Snone -g -C
```
bassosimone, line 56: Can you always keep `5432` here 🙏? Using `15432` does not work on Fedora for me. I suppose it's leaner if we always keep the same port on every possible configuration.

hellais: Yes will do!

bassosimone: 🙏 🙂

FedericoCeratto (approved):

PR #229, opened 2021-03-31
fix(CheckIn): return large list with manual runs

bassosimone:
We want to preserve the behavior where `ooniprobe run` launched
manually from the command line tests the whole test list.

To this end, modify "manual" CheckIn calls to return up to
9,999 URLs, which is what /api/v1/test-list/urls does.

Otherwise, continue implementing the previous behavior where
we return 100 URLs (charging) or 20 URLs (otherwise).

Part of https://github.com/ooni/probe/issues/1299.


hellais (approved):
LGTM

PR #232, opened 2021-04-16
Add authentication and citizenlab API

FedericoCeratto:


newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,213 @@
...
+    return jwt.encode(payload, key, algorithm="HS256").decode()
+
+
+def decode_jwt(token, **kw):
+    key = current_app.config["JWT_ENCRYPTION_KEY"]
+    return jwt.decode(token, key, algorithms=["HS256"], **kw)
+
+
+def hash_email_address(email_address: str) -> str:
+    key = current_app.config["JWT_ENCRYPTION_KEY"].encode()
```
hellais, line 56: Are we sure we want to be using the same key for both the keyed hash function and the signing of the JWT tokens? I haven't looked at it carefully, but I wonder if by doing so we are in some way reducing the security of the key used.

newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,213 @@
...
+    except jwt.exceptions.InvalidSignatureError:
+        return jerror("Invalid credential signature", code=401)
+    except jwt.exceptions.DecodeError:
+        return jerror("Invalid credentials", code=401)
+
+    log.info("user login successful")
+    now = datetime.utcnow()
+    payload = {
+        "nbf": now,
+        "iat": now,
```
hellais, line 205: As discussed we should be adding some sort of expiration time to these tokens to not make them live forever.

hellais:
We discussed this PR during our 1:1. I did another pass at reviewing the code and left just another comment for something to potentially look into.
For the rest this looks great! 🐳

newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,406 @@
...
+    role = _get_account_role(account_id)
+    if role is None:
+        log.info(f"Getting account {account_id} role: not found")
+        return ""
+
+    log.info(f"Getting account {account_id} role: {role}")
+    return role
+
+
+@auth_blueprint.route("/api/v1/set_session_expunge", methods=["POST"])
```
hellais, line 368: `<bikeshed>`
Maybe this API endpoint can be called `/api/v1/refresh_session`?
`</bikeshed>`

newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,406 @@
...
+
+# from ooniapi.utils import cachedjson
+
+auth_blueprint = Blueprint("auth_api", "auth")
+
+"""
+Browser authentication - see probe_services.py for probe authentication
+Requirements:
+  - Never store users email address nor IP addresses nor passwords
+  - Verify email to limit spambots. Do not use capchas
```
hellais, line 27:
```suggestion
  - Verify email to limit spambots. Do not use CAPTCHAs
```

newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,406 @@
...
+        return jerror("Invalid request")
+    if EMAIL_RE.fullmatch(email_address) is None:
+        return jerror("Invalid email address")
+
+    account_id = hash_email_address(email_address)
+    now = datetime.utcnow()
+    expiration = now + timedelta(days=1)
+    # On the backend side the registration is stateless
+    payload = {
+        "nbf": now,
```
hellais, line 193: We might want to account for a small amount of "leeway" to take into account possible issues of invalid tokens due to clock skew (see: https://tools.ietf.org/html/rfc7519#section-4.1.5)
It might be enough to just remove 1 minute from `now`.

newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,406 @@
...
+        "exp": expiration,
```
hellais, line 194: I believe you have to convert these `datetime` objects into an int of seconds since epoch.

I tried running:
```
jwt.encode({'a': datetime.utcnow()}, "antani")
```

and I get
````
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/art/.virtualenvs/ooni-api/lib/python3.7/site-packages/jwt/api_jwt.py", line 60, in encode
    payload, separators=(",", ":"), cls=json_encoder
  File "/Users/art/.pyenv/versions/3.7.3/lib/python3.7/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/Users/art/.pyenv/versions/3.7.3/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/art/.pyenv/versions/3.7.3/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Users/art/.pyenv/versions/3.7.3/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type datetime is not JSON serializable
```

FedericoCeratto: The library handles the special keys by itself:
`jwt.encode({'nbf': datetime.utcnow()}, "antani")`

newapi/ooniapi/citizenlab.py
```diff
@@ -589,3 +590,221 @@ def post_propose_changes():
...
+    sql = """INSERT INTO url_priorities
+        (category_code, cc, domain, url, priority)
+        VALUES(:category_code, '*', '*', '*', :priority)"""
+    category_priorities = {
+        "NEWS": 100,
+        "POLR": 100,
+        "HUMR": 100,
+        "LGBT": 100,
+        "ANON": 100,
+        "GRP": 80,
```
hellais, line 633: We want this to be more highly prioritised than NEW. We should bump it up to 150.

newapi/ooniapi/app.py
```diff
@@ -20,9 +20,9 @@
 
 from flasgger import Swagger
 
-from flask_mail import Mail  # debdeps: python3-flask-mail
+#from flask_mail import Mail  # debdeps: python3-flask-mail
 
-from flask_security import Security  # debdeps: python3-flask-security
+#from flask_security import Security  # debdeps: python3-flask-security
```
hellais, line 25: Perhaps remove these lines if these libraries are no longer needed.

newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,411 @@
...
+        roles = [roles, ]
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            token = request.cookies.get("ooni", "")
+            try:
+                dec = decode_jwt(token, audience="user_auth")
+                if dec["role"] not in roles:
+                    return jerror("Role not authorized", 401)
+            except Exception:
```
hellais, line 74: Can we trap the specific exception that is a token validation issue, instead of the broad `Exception` type. I am worried this might ghost potential bugs in the decoding logic.

newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,465 @@
...
+            # run the HTTP route method
+            resp = func(*args, **kwargs)
+
+            token_age = time.time() - tok["iat"]
+            if token_age > 600:  # refresh token if needed
+                newtoken = _create_session_token(
+                    tok["account_id"], tok["nick"], tok["role"], tok["login_time"]
+                )
+                resp.set_cookie(
+                    "ooni", newtoken, secure=True, httponly=True, samesite="Strict"
```
hellais, line 120: Perhaps this can be made into a function for setting the cookie so that we can be sure we always set the cookies in the same way?

newapi/ooniapi/auth.py
hellais: Actually this comment is not really so much relevant for this section where we send the email. It's more relevant for the other places in which a token that needs to be used via cookies is issued.

newapi/ooniapi/auth.py
```diff
@@ -0,0 +1,465 @@
...
+            email_address:
+              type: string
+    responses:
+      200:
+        description: Confirmation
+    """
+    log = current_app.logger
+    req = request.json if request.is_json else request.form
+    nick = req.get("nickname", "").strip()
+    # Accept all alphanum including unicode and whitespaces
```
hellais, line 208: Are we sure we want spaces in nicknames?

hellais (approved):
I did a pass at reviewing this and it looks good to me. I left some comments for improvements or questions that I had while reading it. They can be addressed either now or in future work.

I don't see this in blocking on landing this!

newapi/ooniapi/auth.py
FedericoCeratto: Done

newapi/ooniapi/auth.py
FedericoCeratto: Done

newapi/ooniapi/auth.py
FedericoCeratto: Done

FedericoCeratto:
0.26~pr232-513 deployed on -test

newapi/ooniapi/citizenlab.py
```diff
@@ -0,0 +1,822 @@
...
+    """
+    global log
+    log = current_app.logger
+    username = get_username()
+    ulm = get_url_list_manager()
+    tl = ulm.get_test_list(username, country_code)
+    return make_response(jsonify(tl))
+
+
+@cz_blueprint.route("/api/v1/url-submission/add-url", methods=["POST"])
```
hellais, line 464: Perhaps for consistency with the `/api/_/url-submission` endpoint we should do the add-url with the same pattern of having only a `new_entry` an no `old_entry`

newapi/ooniapi/citizenlab.py
```diff
@@ -0,0 +1,822 @@
...
+@role_required(["admin", "user"])
+def url_submission_add_url():
```
hellais, line 466: It seems like we are not checking for duplicate URLs in subsequent submissions to this API endpoint. See: https://github.com/citizenlab/test-lists/pull/773.

Steps to reproduce:

```
r = s.post("https://ams-pg-test.ooni.org/api/v1/url-submission/add-url", json={
    "country_code": "IT",
    "comment": "Add thepiratebay.se to the italy list",
    "new_entry": [
        "https://thepiratebay.se/",
        "FILE",
        "File-sharing",
        "2021-05-18",
        "arturo",
        "Site is blocked"
    ]
})
r = s.post("https://ams-pg-test.ooni.org/api/v1/url-submission/add-url", json={
    "country_code": "IT",
    "comment": "Add thepiratebay.se to the italy list",
    "new_entry": [
        "https://thepiratebay.se/",
        "FILE",
        "File-sharing",
        "2021-05-18",
        "arturo",
        "Site is blocked"
    ]
})
r.json()
```

newapi/ooniapi/auth.py
FedericoCeratto: The refreshing is done by the browser of the user. This entry point is used by admins to flag sessions to be expunged and force a refresh if/when the user comes back.

newapi/ooniapi/auth.py
FedericoCeratto: The timestamping and verification is done by the same host.

newapi/ooniapi/app.py
FedericoCeratto: Done

PR #236, opened 2021-05-10
0.28 serve Tor and Psiphon configs

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/493

### DIFF
--- a/newapi/ooniapi/probe_services.py
+++ b/newapi/ooniapi/probe_services.py
@@ -175,12 +175,20 @@ def check_in():
         test_items = []
 
     metrics.gauge("check-in-test-list-count", len(test_items))
+    try:
+        torconf = _load_json(current_app.config["TOR_TARGETS_CONFFILE"])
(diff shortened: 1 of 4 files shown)


newapi/ooniapi/probe_services.py
```diff
@@ -171,12 +171,20 @@ def check_in():
...
+        log.error(str(e), exc_info=1)
+        conf = {}
+
     resp = dict(
         v=1,
         tests={
             "web_connectivity": {"urls": test_items},
         },
-        conf={},
+        conf=conf,
```
hellais, line 187: Shouldn't this go nested inside of the `tests` key under the testname it's relevant to?

FedericoCeratto: `tests` contains data about the tests to be run right now and should not be cached on the probes

`conf` contains more persistent configuration and should be cached by the probes

PR #238, opened 2021-05-10
Switch packaging to S3

FedericoCeratto:


.circleci/config.yml
```diff
@@ -42,5 +46,6 @@ workflows:
 
   my_workflows:
     jobs:
-      - create_deb_package
-      - end_to_end_test
+      - create_deb_package:
+          context: ooni
+      #- end_to_end_test
```
hellais, line 51: Did you comment out the end_to_end test because it wasn't working?

I had similar issues getting it to work inside of: https://github.com/ooni/api/pull/240

hellais (commented):
Maybe we should consolidate the efforts of this PR together with those of https://github.com/ooni/api/pull/240.

It seems like the direction we are going is having things run on github actions, so maybe it's optimal to go for landing #240 instead of this one?

hellais:
Can we close this PR?

PR #240, opened 2021-05-14
Fix debian repo URL

hellais:


hellais:
This is superseeded by https://github.com/ooni/api/pull/244

PR #244, opened 2021-05-21
Docker compose tests

hellais:
This PR achieves the following:
* Switch to s3 for the hosting of the debian builds
* Port the workflow for building packages over to github actions
* Create a docker-compose based system for running tests on developers laptop and on the CI

This is based on top of https://github.com/ooni/api/pull/240


newapi/api.conf.example
```diff
@@ -23,11 +23,11 @@ TOR_TARGETS_CONFFILE = "/etc/ooni/tor_targets.json"
 # Registration email delivery
 MAIL_SERVER = "CHANGEME"
 MAIL_PORT = 465
-MAIL_USE_SSL = True
+MAIL_USE_SSL = False
 MAIL_USERNAME = "CHANGEME"
 MAIL_PASSWORD = "CHANGEME"
 MAIL_SOURCE_ADDRESS = "contact@ooni.org"
-LOGIN_BASE_URL = "https://CHANGEME/login"
+LOGIN_BASE_URL = "https://api.ooni.io/api/v1/user_login"
```
FedericoCeratto, line 30: This is an example for 3rd parties: we should not hardcode parameters related to the ooni infra.

hellais: It's a dependency for the tests to run properly. Should I split out the config file used for testing from the config file used as example?

.github/workflows/mypy.yml
```diff
@@ -13,14 +13,16 @@ jobs:
       - name: Setup APT
         run: |
           apt-get update
-          apt-get install --no-install-recommends -y ca-certificates
+          apt-get install --no-install-recommends -y ca-certificates gnupg
```
FedericoCeratto, line 16: Do we need gnupg here?

hellais: Don't we need it to install the GPG key of the deb-ci repo?

FedericoCeratto: no, that's done by apt itself

newapi/api.conf.example
FedericoCeratto: Yes: the example for 3rd party users should have "CHANGEME"s to indicate  what needs customization. The files for local integ test and github CI have specific values for those envs.

FedericoCeratto (changes requested):
The CI run in github is going to become very complex especially once we start running the fastpath in it. We need to ensure that tmate works reliably in order to be able to debug and maintain it.

hellais:
Added support for manually triggering tmate runs

FedericoCeratto (approved):

PR #247, opened 2021-05-31
0.31 Fix CORS

FedericoCeratto:


sarathms (approved):
👍  This seems to work. I wonder if it would be useful to configure it as an environment variable so we can add/remove entries as we deploy or take down those services.

PR #250, opened 2021-06-14
Handle reprocessed msmts in jsonl table

FedericoCeratto:
Introduce looking up msmts by measurement_uid

### DIFF
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -8,7 +8,7 @@ from datetime import datetime, timedelta, date
 from dateutil.parser import parse as parse_date
 from io import StringIO
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Optional
 import gzip
 import http.client
 import json
@@ -27,7 +27,7 @@ from flask import current_app, request, make_response, abort, redirect
 from flask.json import jsonify
 from werkzeug.exceptions import HTTPException, BadRequest
 
-from sqlalchemy import func, and_, false, text, select, sql, column
+from sqlalchemy import func, and_, text, select, sql, column
 from sqlalchemy.sql import literal_column
 from sqlalchemy import String, cast
 from sqlalchemy.exc import OperationalError
@@ -37,7 +37,6 @@ from urllib.request import urlopen
 from urllib.parse import urljoin, urlencode
 
 from ooniapi import __version__
-from ooniapi.config import REPORT_INDEX_OFFSET
 from ooniapi.config import metrics
 from ooniapi.utils import cachedjson
 from ooniapi.models import TEST_NAMES
@@ -58,6 +57,10 @@ log = logging.getLogger()
 class QueryTimeoutError(HTTPException):
     code = 504
     description = "The database query timed out.\nTry changing the query parameters."
+
+
+class MsmtNotFound(Exception):
+    pass
 
 
 def get_version():
@@ -173,7 +176,6 @@ def _fetch_jsonl_measurement_body_inner(
     linenum: int,
 ) -> bytes:
     log = current_app.logger
-    REQID_HDR = "X-Request-ID"
     # TODO configure from file
     BASEURL = "https://ooni-data-eu-fra.s3.amazonaws.com/"
     url = urljoin(BASEURL, s3path)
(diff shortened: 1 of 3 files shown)


newapi/ooniapi/measurements.py
```diff
@@ -184,32 +186,39 @@ def _fetch_jsonl_measurement_body_inner(
...
-        input = ""
-    query_params = dict(input=input, report_id=report_id)
+    query = "SELECT s3path, linenum FROM jsonl "
+    if measurement_uid is None:
+        query += "WHERE report_id = :report_id AND input = :input LIMIT 1"
+        query_params = dict(input=input, report_id=report_id)
+
+    else:
+        query += """WHERE measurement_uid = :mid
+        OR (report_id = :rid AND input = :inp)
```
hellais, line 201: Why do we need the OR here? If the `measurement_uid` has been passed, shouldn't it always return a match?

newapi/ooniapi/measurements.py
```diff
@@ -284,38 +293,39 @@ def _fetch_autoclaved_measurement_body(report_id: str, input) -> Optional[bytes]
...
-    # The number of underscores can be:
-    # 1: Legacy measurement e.g.
-    # 20141101T220015Z_OzDkiPoJMVjHItj<redacted>
-    # 2: Legacy measurement e.g.
-    # 20201019T050625Z_AS8369_VVqkepmw<redacted>
     # 5: Current format e.g.
     # 20210124T210009Z_webconnectivity_VE_22313_n1_Ojb<redacted>
     if u_count == 5:
         # Look on disk and then from JSONL cans on S3
         body = _fetch_measurement_body_on_disk(report_id, input)
```
hellais, line 312: Maybe for consistency this function should also `raise MsmtNotFound` instead of checking for the body being None.

newapi/ooniapi/measurements.py
```diff
@@ -284,38 +293,39 @@ def _fetch_autoclaved_measurement_body(report_id: str, input) -> Optional[bytes]
...
-        if body is None:
-            log.debug(f"Fetching body for {report_id} {input} from jsonl on S3")
-            body = _fetch_jsonl_measurement_body(report_id, input)
-    elif u_count == 2:
-        body = _fetch_autoclaved_measurement_body(report_id, input)
-    elif u_count == 1:
-        body = _fetch_autoclaved_measurement_body(report_id, input)
-    else:
-        raise BadRequest("Invalid report_id")
-    return body
```
hellais, line 318: I like how this is significantly simplified now! 👍

newapi/ooniapi/measurements.py
```diff
@@ -484,7 +495,14 @@ def get_measurement_meta():
...
         return cachedjson(24, **msmt_meta)
 
-    body = _fetch_measurement_body(report_id, input)
+    try:
+        body = _fetch_measurement_body(report_id, input, msmt_meta["measurement_uid"])
+        assert isinstance(body, bytes)
+        body = body.decode()
+    except Exception as e:
+        log.error(e, exc_info=1)
+        body = ""
```
hellais, line 504: Shouldn't we raise an exception rather than returning an empty body here?

newapi/ooniapi/measurements.py
FedericoCeratto: No, we don't have uids for most msmts

newapi/ooniapi/measurements.py
FedericoCeratto: No, it's what Explorer expects

newapi/ooniapi/measurements.py
FedericoCeratto: Not finding the msmt on disk is not an error. It's expected to happen often.

newapi/ooniapi/measurements.py
hellais: Even after reprocessing all of the legacy ones?

FedericoCeratto:
Deployed on -test using reprocessed data.
See: https://explorer-test.ooni.io/search?until=2018-11-1&since=2018-10-25

newapi/ooniapi/measurements.py
FedericoCeratto: Yes, they are the ones with uid. The new ones don't have it.

hellais: Ah right, so until we reprocess also the new data, we won't have the measurement_uid in the jsonl table, but only inside of the fastpath table. Is this correct?

FedericoCeratto: Yes. After we reprocess everything we'll be able to simplify the query, also drop the autoclaved-related queries, and remove a DB index.

PR #252, opened 2021-06-15
Citizenlab tweaks

FedericoCeratto:


newapi/ooniapi/citizenlab.py
```diff
@@ -282,12 +282,19 @@ def add(self, username, cc, new_entry, comment):
...
+        if old_entry:
+            old_entry_li = [old_entry[k] for k in ks]
+        if new_entry:
+            new_entry_li = [new_entry[k] for k in ks]
         log.debug("updating existing entry")
         cc = cc.lower()
         if len(cc) != 2:
             raise Exception("Invalid country code")
 
+        #entry["category_desc"] = CATEGORY_CODES[entry["category_code"]]
```
hellais, line 296: I think this is a good idea to ensure they are always consistent.

hellais: Maybe we should do the same for `date_str`

newapi/ooniapi/citizenlab.py
```diff
@@ -282,12 +282,19 @@ def add(self, username, cc, new_entry, comment):
 
             self.set_state(username, "IN_PROGRESS")
 
-    def update(self, username, cc, old_entry, new_entry, comment):
+    def update(self, username: str, cc: str, old_entry: dict, new_entry: dict, comment):
+        ks = ("url", "category_code", "category_desc", "date_str", "user", "notes")
```
hellais, line 286: The column is called `date_added` not `date_str`

hellais: It's also called `category_description` instead of `category_desc`

hellais: `user` should be `source`

newapi/ooniapi/citizenlab.py
```diff
@@ -488,7 +497,17 @@ def url_submission_add_url():
...
               type: string
             new_entry:
-              type: array
+              type: object
+              properties:
+                category_code:
+                  type: string
+                url:
+                  type: string
+                date_str:
```
hellais, line 506: It's called `date_added` in the citizenlab test-lists repo.

newapi/ooniapi/citizenlab.py
FedericoCeratto: We discussed deduplicating this field.

newapi/ooniapi/citizenlab.py
```diff
@@ -535,10 +554,32 @@ def url_submission_update_url():
...
+                  type: string
+                user:
+                  type: string
```
hellais, line 580: See above comments for the naming of the columns.

newapi/ooniapi/citizenlab.py
```diff
@@ -411,14 +422,12 @@ def check_url(url):
...
 
-def validate_entry(entry: List[str]) -> None:
-    url, category_code, category_desc, date_str, user, notes = entry
-    check_url(url)
-    if category_code not in CATEGORY_CODES:
+def validate_entry(entry: Dict[str, str]) -> None:
+    check_url(entry["url"])
+    if entry["category_code"] not in CATEGORY_CODES:
         raise BadCategoryCode()
-    if category_desc != CATEGORY_CODES[category_code]:
```
hellais, line 419: You should either leave this check of the category_desc, or uncomment the line above where you override it.

newapi/ooniapi/citizenlab.py
FedericoCeratto: Fixed

newapi/ooniapi/citizenlab.py
hellais: Yes I agree with setting it in the backend, I am saying that maybe we can also rely on the backend for stamping the time of when a URL was added/edited.

newapi/ooniapi/citizenlab.py
FedericoCeratto: (I'm aware, the code is WIP)

newapi/ooniapi/citizenlab.py
```diff
@@ -535,10 +486,32 @@ def url_submission_update_url():
...
+                  type: string
+                url:
+                  type: string
+                date_added:
+                  type: string
+                user:
+                  type: string
+                notes:
+                  type: string
+            new_entry:
```
sarathms, line 502: I think this also implies that the type for `updated_entry` in the response should also be changed to `object`.

newapi/ooniapi/citizenlab.py
FedericoCeratto: Renamed

newapi/ooniapi/citizenlab.py
FedericoCeratto: Updated. Is it useful or should we drop it?

newapi/ooniapi/citizenlab.py
FedericoCeratto: Added a TODO

newapi/ooniapi/citizenlab.py
```diff
@@ -411,16 +409,17 @@ def check_url(url):
         raise BadURL()
 
 
-def validate_entry(entry: List[str]) -> None:
-    url, category_code, category_desc, date_str, user, notes = entry
-    check_url(url)
-    if category_code not in CATEGORY_CODES:
+def validate_entry(entry: Dict[str, str]) -> None:
+    keys = ["category_code", "date_added", "notes", "url", "user"]
```
hellais, line 413: They `user` key should be called `source`

newapi/ooniapi/citizenlab.py
hellais: Should be `source` instead of `user`.

newapi/ooniapi/citizenlab.py
```diff
@@ -254,36 +254,28 @@ def sync_state(self, username):
...
-        self.sync_state(username)
-        self.pull_origin_repo()
-        log.debug("adding new entry")
-        state = self.get_state(username)
-        if state in ("PR_OPEN"):
-            raise Exception("You cannot edit files while changes are pending")
+    def update(self, username: str, cc: str, old_entry: dict, new_entry: dict, comment):
+        """Create/update/delete"""
+        # TODO: set date_added to now() on new_entry
+        ks = ("url", "category_code", "date_added", "source", "notes")
```
hellais, line 260: You are missing the `category_description` from the `old_entry_li` so below when you check for `row == old_entry_li` it will always result in a miss.

newapi/ooniapi/citizenlab.py
FedericoCeratto: Commented out for now

newapi/ooniapi/citizenlab.py
FedericoCeratto: Updated

PR #253, opened 2021-06-21
Configure CORS

FedericoCeratto:


newapi/ooniapi/app.py
```diff
@@ -120,20 +121,12 @@ def init_app(app, testmode=False):
 
     # FIXME
     # CORS(app, resources={r"/api/*": {"origins": "*"}})
-    # TODO: configure from api.conf:
     orig = [
-        "https://explorer-test.ooni.io",
-        "https://explorer-test.ooni.org",
-        "https://explorer.ooni.org",
-        "https://url-prioritization.ooni.org",
```
sarathms, line 128: Will the environment variable `CORS_URL` hold these URLs in it (and maybe more or less as we need)?
The name `CORS_URL` indicates it will be just one URL which, if it is something like `https://api.ooni.io` like in `api.config.example`, might not serve the purpose.

newapi/ooniapi/app.py
```diff
@@ -56,6 +56,7 @@ def validate_conf(app, conffile):
     conf_keys = (
         "BASE_URL",
         "COLLECTORS",
+        "CORS_URL",
```
sarathms, line 59:
```suggestion
        "CORS_URLS",
```

sarathms (approved):
LGTM 👍

hellais (approved):
Maybe remove the CORS_URLS settings or add a comment that it's unused.

PR #258, opened 2021-07-15
Implement end-to-end CI

FedericoCeratto:


newapi/tests/conftest.py
```diff
@@ -53,7 +60,121 @@ def client(app):
...
+    cmd = cmd.split()
+    subprocess.run(cmd, check=True, stdout=PIPE, stderr=PIPE).stdout
+    return Path(d)
+
+
+def run_sql_scripts(app):
+    log = app.logger
+    # for i in ["1_metadb_users.sql", "2_metadb_schema.sql", "3_test_fixtures.sql"]:
+    query = ""
+    for i in ["2_metadb_schema.sql", "3_test_fixtures.sql"]:
```
hellais, line 120: These are already run as part of the docker-compose db definition. If you decide to switch to the pattern of running them in here, you probably should remove them from: https://github.com/ooni/api/pull/258/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R12

newapi/tests/conftest.py
```diff
@@ -53,7 +60,121 @@ def client(app):
...
+@pytest.fixture(scope="session")
+def setup_database_part_1():
+    # Create database and users.
+    # Drop and recreate database if exists.
+    if not pytest.create_db:
+        return
+
+    if os.path.exists("/usr/bin/sudo"):
+        print("Creating PostgreSQL user and database")
+        sudopg("DROP DATABASE IF EXISTS oonitestdb", check=True)
```
hellais, line 91: I suspect these will not work because you are running them from inside of the ooni_api container instead of the DB container. To target the `db` container you will have to use network based authentication rather than socket based.

newapi/tests/conftest.py
```diff
@@ -53,7 +60,121 @@ def client(app):
...
+        sudopg("DROP ROLE IF EXISTS oonitest", check=True)
+        sudopg("CREATE USER oonitest WITH ENCRYPTED PASSWORD 'test'", check=False)
+        sudopg("CREATE DATABASE oonitestdb WITH OWNER 'oonitest'", check=False)
+        sudopg("GRANT ALL PRIVILEGES ON DATABASE oonitestdb TO oonitest")
```
hellais, line 95: Isn't this already done over here: https://github.com/ooni/api/pull/258/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R9?

hellais (commented):
I left some comments for things to think about. I also see that the tests are currently failing due to the `db` host not being reachable. This is most likely caused by the fact that the postgresql docker-compose image failed to be created.

newapi/tests/conftest.py
FedericoCeratto: "sudo" is being used or not based on the environment in order to be able to run the tests in different environments. See the comment " On github sudo is missing and the database is already created "

newapi/tests/conftest.py
FedericoCeratto: See previous comment

newapi/tests/conftest.py
FedericoCeratto: See previous comment.

PR #263, opened 2021-08-09
List/fetch measurements by UID

FedericoCeratto:


newapi/ooniapi/measurements.py
```diff
@@ -228,6 +228,26 @@ def _unwrap_post(post: dict) -> dict:
     raise Exception("Unexpected format")
 
 
+def _read_measurement_from_disk(msmt_uid: str) -> Optional[bytes]:
+    """Reads msmt from local disk cache or returns None"""
+    assert msmt_uid
+    if not msmt_uid.startswith("20"):
```
sarathms, line 234: I think our this tool is solid enough to be around till `2099-12-31` 😜

newapi/ooniapi/measurements.py
```diff
@@ -59,8 +59,9 @@ class QueryTimeoutError(HTTPException):
     description = "The database query timed out.\nTry changing the query parameters."
 
 
-class MsmtNotFound(Exception):
-    pass
+class MsmtNotFound(HTTPException):
+    code = 500
```
sarathms, line 63: Why `500` and not `404`?

FedericoCeratto: The HTTP route is found. The measurement is expected to be found and this constitutes an error on the server side.

newapi/ooniapi/measurements.py
```diff
@@ -228,6 +239,26 @@ def _unwrap_post(post: dict) -> dict:
     raise Exception("Unexpected format")
 
 
+def _read_measurement_from_disk(msmt_uid: str) -> Optional[bytes]:
+    """Reads msmt from local disk cache or returns None"""
+    assert msmt_uid
+    if not msmt_uid.startswith("20"):
+        return None
```
hellais, line 246: Isn't it better to assert this and raise, rather than returning a `None` value?

newapi/ooniapi/measurements.py
```diff
@@ -337,7 +367,7 @@ def genurl(path: str, **kw) -> str:
 @api_msm_blueprint.route("/v1/raw_measurement")
 @metrics.timer("get_raw_measurement")
 def get_raw_measurement():
-    """Get raw measurement body by measurement_id + input
+    """Get raw measurement body by either measurement_id + input or measurement_uid
```
hellais, line 370:
```suggestion
    """Get raw measurement body by either report_id + input or measurement_uid
```

hellais (approved):
Left few very minor comments, otherwise looks good to merge.

FedericoCeratto:
Moved to backend repo.

newapi/ooniapi/measurements.py
FedericoCeratto: It's not an exception because it's expected to happen often.

PR #264, opened 2021-08-26
0.44 Fix prioritization

FedericoCeratto:
- Bugfix
- Add ability to disable an URL or category completely by setting a very low negative priority
- Improve testing


newapi/ooniapi/prio.py
```diff
@@ -133,16 +141,41 @@ def fetch_reactive_url_list(cc: str):
...
     q = current_app.db_session.execute(sql, dict(cc=cc, cc_low=cc.lower()))
     entries = tuple(q.fetchall())
     log.info("%d entries", len(entries))
-    return entries
+
+    # Fetch url priorities rules
+    sql = """SELECT category_code, cc, domain, url, priority
+    FROM url_priorities WHERE cc = :cc OR cc = '*'
+    """
+    q = current_app.db_session.execute(sql, dict(cc=cc))
```
hellais, line 154: Since you support upper and lowercase match in the citizenlab table, perhaps you should give a similar treatment to the cc match in the `url_priorities` table?

newapi/tests/integ/test_prioritization_nodb.py
```diff
@@ -0,0 +1,136 @@
...
+    ): Path("citizenlab.json"),
+    (
+        "SELECT category_code, domain, url, cc, COALESCE(msmt_cnt, 0)::float AS msmt_cnt FROM ( SELECT domain, url, cc, category_code FROM citizenlab WHERE citizenlab.cc = :cc_low OR citizenlab.cc = :cc OR citizenlab.cc = 'ZZ' ) AS citiz LEFT OUTER JOIN ( SELECT input, msmt_cnt FROM counters_test_list WHERE probe_cc = :cc ) AS cnt ON (citiz.url = cnt.input)"
+    ): Path("citizenlab_counters_us.json"),
+    (
+        "SELECT category_code, cc, domain, url, priority FROM url_priorities WHERE cc = :cc OR cc = '*'"
+    ): Path("url_priorities_us.json"),
+}
+
+
```
hellais, line 30: Since these tests are based on deterministic database dumps, perhaps we can also add a check which verifies that the sorting of the returned URLs matches the expected result (i.e. the priorities are in fact being respected)?

hellais (approved):
I reviewed the diff and it looks good to me. I left a couple of comments for improvements that can either be done as part of this PR or as future work.

newapi/ooniapi/prio.py
FedericoCeratto: No, url_priorities is correctly using uppercase. We might want to switch citizenlab to uppercase in future.

PR #265, opened 2021-09-03
0.45 check-in and priority improvements

FedericoCeratto:
- Sort URL prioritization rules
- Cleanup
- Add tests
- Return UTC time in check-in call


hellais (approved):
LGTM

newapi/ooniapi/probe_services.py
```diff
@@ -183,12 +186,14 @@ def check_in():
         log.error(str(e), exc_info=1)
         conf = {}
 
+    utc_time = datetime.utcnow().strftime('%Y%m%dT%H%M%S')
```
hellais, line 189: The format of this timestamp does not match the docstring. I would suggest, for consistency with other places where we have timestamp, to use the format: `YYYY-mm-ddTHH:MM:SSZ`

FedericoCeratto: Fixed

PR #270, opened 2021-11-15
0.49 Fix logic of citizenlab URL submission backend

hellais:
Previously it would not work properly when configured in our testing vs
production setup as several assumptions were made about the push and
origin repos.

Specifically:

1. The github_username was being derived from the push_repo address,
which might not be a username in the case of a repo configured under a
github org

2. The address used for opening pull requests was the push_repo, it
should instead be the origin_repo

### DIFF
--- a/newapi/ooniapi/citizenlab.py
+++ b/newapi/ooniapi/citizenlab.py
@@ -190,7 +190,7 @@ class URLListManager:
         """
         apiurl = self.get_pr_id(account_id)
         pr_num = apiurl.split("/")[-1]
-        return f"https://github.com/{self.push_repo}/pull/{pr_num}"
+        return f"https://github.com/{self.origin_repo}/pull/{pr_num}"
 
     def get_user_repo(self, account_id: str):
         repo_path = self.get_user_repo_path(account_id)
--- a/newapi/ooniapi/citizenlab.py
+++ b/newapi/ooniapi/citizenlab.py
@@ -368,7 +368,7 @@ class URLListManager:
         https://api.github.com/repos/citizenlab/test-lists/pulls/800
         """
         head = f"{self.push_username}:{branchname}"
-        log.info(f"opening a PR for {head} on {self.origin_repo}")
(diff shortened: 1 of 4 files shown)


newapi/ooniapi/citizenlab.py
```diff
@@ -98,13 +98,14 @@ def update(self, op_code, cur_count, max_count=None, message=""):
...
-    def __init__(self, working_dir, github_token, push_repo, origin_repo):
+    def __init__(self, working_dir, github_user, github_token, push_repo, origin_repo):
         self.working_dir = working_dir
         self.origin_repo = origin_repo
         self.push_repo = push_repo
-        self.github_user = push_repo.split("/")[0]
+        self.github_user = github_user
         self.github_token = github_token
         self.repo_dir = self.working_dir / "test-lists"
+        self.push_username = push_repo.split("/")[0]
```
FedericoCeratto, line 108: Maybe this should be renamed push_orgname?

newapi/ooniapi/citizenlab.py
```diff
@@ -366,17 +367,17 @@ def open_pr(self, branchname):
         """Opens PR. Returns API URL e.g.
         https://api.github.com/repos/citizenlab/test-lists/pulls/800
         """
-        head = f"{self.github_user}:{branchname}"
-        log.info(f"opening a PR for {head} on {self.push_repo}")
+        head = f"{self.push_username}:{branchname}"
+        log.info(f"opening a PR for {head} on {self.origin_repo}")
```
FedericoCeratto, line 371: log.info(f"opening a PR for {head} on {self.origin_repo} using {self.push_repo}")

newapi/ooniapi/citizenlab.py
hellais: It can be either an org or a username, so it's kind of equivalent which one we choose. I picked the `username` naming for consistency with how it's documented in the github API docs: https://docs.github.com/en/rest/reference/pulls#create-a-pull-request

PR #271, opened 2021-11-18
Add Clickhouse support

FedericoCeratto:


newapi/ooniapi/measurements.py
```diff
@@ -1062,45 +1514,103 @@ def get_aggregated():
...
-        coalsum("failure_count"),
-        coalsum("measurement_count"),
+        sql.text("SUM(anomaly = 't') AS anomaly_count"),
+        sql.text("SUM(confirmed = 't') AS confirmed_count"),
+        sql.text("SUM(msm_failure = 't') AS failure_count"),
+        sql.text("COUNT(*) AS measurement_count"),
     ]
-    table = sql.table("counters")
+    # table = sql.table("counters")
+    table = sql.table("fastpath")
```
hellais, line 1603: Why has the aggregation function for using the clickhouse backend been changed to use the fastpath instead of the counters tables?

newapi/ooniapi/measurements.py
```diff
@@ -1062,45 +1514,103 @@ def get_aggregated():
...
+    axis_x,
+    axis_y,
+):
+
+    dimension_cnt = int(bool(axis_x)) + int(bool(axis_y))
+    cacheable = until and until < datetime.now() - timedelta(hours=72)
 
     # Assemble query
     def coalsum(name):
         return sql.text("COALESCE(SUM({0}), 0) AS {0}".format(name))
```
hellais, line 1594: Maybe then `coalsum` can be deleted?

newapi/ooniapi/measurements.py
```diff
@@ -1198,6 +1708,175 @@ def coalsum(name):
...
+    probe_cc,
+    probe_asn,
+    test_name,
+    axis_x,
+    axis_y,
+):
+    log = current_app.logger
+    dimension_cnt = int(bool(axis_x)) + int(bool(axis_y))
+    cacheable = until and until < datetime.now() - timedelta(hours=72)
+    cacheable = False  # FIXME
```
hellais, line 1727: Since measurements can also be uploaded at a later stage I recon we should disable caching altogether on these endpoints.

newapi/ooniapi/measurements.py
```diff
@@ -1198,6 +1708,175 @@ def coalsum(name):
...
+        where.append(sql.text("category_code = :category_code"))
+        query_params["category_code"] = category_code
+        log.error("@"*66)
+
+    if probe_cc:
+        where.append(sql.text("probe_cc = :probe_cc"))
+        query_params["probe_cc"] = probe_cc
+
+    if probe_asn is not None:
+        if probe_asn.startswith("AS"):
```
hellais, line 1764: Isn't this already handled earlier on by the `param_asn` function?

newapi/ooniapi/measurements.py
```diff
@@ -1198,6 +1708,175 @@ def coalsum(name):
...
+    if since:
+        where.append(sql.text("measurement_start_time > :since"))
+        query_params["since"] = since
+
+    if until:
+        where.append(sql.text("measurement_start_time <= :until"))
+        query_params["until"] = until
+
+    if test_name:
+        assert test_name in TEST_NAMES
```
hellais, line 1779: Perhaps this should be handled in the parent function and a user meaningful error should be returned in the event of an unsupported test_name.

newapi/ooniapi/measurements.py
```diff
@@ -1198,6 +1708,175 @@ def coalsum(name):
...
+            delta = 0
+
+        # lim = 100
+        # if delta > lim * 30:
+        #     granularity = ("year", "toYear")
+        # elif delta > lim * 7:
+        #     granularity = ("month", "toMonth")
+        # elif delta > lim:
+        #     granularity = ("week", "toWeek")
+        if delta < 2:
```
hellais, line 1802: I am not sure it's a good idea for the backend to be doing this type of dynamic/magic scaling of the granularity of the returned data. I would rather it's the user explicitly asking for a higher or lower level of granularity and them passing it to the backend via the UI.

newapi/ooniapi/measurements.py
```diff
@@ -1198,6 +1708,175 @@ def coalsum(name):
...
+            granularity = ("day", "toDate")
+        tcol, fun = granularity
+        tcol = "measurement_start_day"  # TODO: support dynamic axis names
+        cols.append(sql.text(f"{fun}(measurement_start_time) AS {tcol}"))
+        colnames.append(tcol)
+        group_by.append(column(tcol))
+
+    def add_axis(axis, cols, colnames, group_by):
+        if axis == "blocking_type":
+            t = "JSONExtractString(scores, 'analysis', 'blocking_type') AS blocking_type"
```
hellais, line 1814: Nice!

newapi/ooniapi/measurements.py
```diff
@@ -719,29 +836,82 @@ def list_measurements():
...
+    failure,
+    input_,
+    domain,
+    category_code,
+    order,
+    order_by,
+    limit,
+    offset,
+):
     INULL = ""  # Special value for input = NULL to merge rows with FULL OUTER JOIN
```
hellais, line 910: This can be deleted now I guess.

newapi/ooniapi/measurements.py
```diff
@@ -865,6 +1031,8 @@ def list_measurements():
     # Assemble the "external" query. Run a final order by followed by limit and
     # offset
     query = fp_query.offset(offset).limit(limit)
+    query_params["param_1"] = limit
+    query_params["param_2"] = offset
```
hellais, line 1035: Where are these being used?

newapi/ooniapi/measurements.py
FedericoCeratto: No, why? It's the Postgres version of the query.

newapi/ooniapi/measurements.py
FedericoCeratto: No, this is the postgresql version of the query.

newapi/ooniapi/measurements.py
FedericoCeratto: IIRC it was required by offset(...) and limit(...) but I'm not focusing on PG code in this PR.

newapi/ooniapi/measurements.py
FedericoCeratto: This is why is mostly commented out. Users might be unaware of what constitutes the right granularity and poor choices impact the API; also we might want to be able to tune the granularity centrally over time to optimize performance.

However we could let the API pick the preferred granularity by default, to make the MAT user friendly, and also allow users to require a fixed granularity. In the second case the API could simply refuse overly heavy queries.

newapi/ooniapi/measurements.py
hellais: I didn't notice that it was being used further down, I thought it was no longer used since it was dropped from the query itself.

newapi/ooniapi/citizenlab.py
```diff
@@ -750,13 +753,110 @@ def list_url_priorities():
...
+        ('MISC', '*', '*', '*', 20),
+        ('HACK', '*', '*', '*', 20),
+        ('ECON', '*', '*', '*', 20),
+        ('COMM', '*', '*', '*', 20),
+        ('CTRL', '*', '*', '*', 20),
+        ('NEWS', 'it', '*', '*', 10),
+        ('NEWS', 'it', 'www.leggo.it', '*', 5),
+        ('COMT', '*', '*', '*', 100),
+        ('GRP', '*', '*', '*', 100),
+    ]
```
hellais, line 835: Why are these priorities hardcoded?

How will the updating of the priorities work? Will the prioritization frontend write to pg or clickhouse? How do we ensure the two tables are kept in sync? If it's not needed to keep them in sync, perhaps we should be dropping support for reading/writing to whichever one is no longer being used.

newapi/ooniapi/measurements.py
FedericoCeratto: Fixed

newapi/ooniapi/measurements.py
FedericoCeratto: The caching is temporarily disabled but could be added back and used only when the msmts are old enough to safely support caching.

newapi/ooniapi/measurements.py
FedericoCeratto: Done

newapi/ooniapi/citizenlab.py
FedericoCeratto: This was used only once just to bootstrap default values in the table.

newapi/tests/conftest.py
```diff
@@ -101,20 +108,22 @@ def setup_database_part_1():
...
+    """Clone pipeline repo to then run fastpath from S3 and citizenlab importer"""
+    if not pytest.create_db and not pytest.inject_msmts:
         return
     d = tmpdir_factory.mktemp("pipeline")
     if d.isdir():
         shutil.rmtree(d)
     # cmd = f"git clone --depth 1 https://github.com/ooni/pipeline -q {d}"
-    cmd = f"git clone --depth 1 https://github.com/ooni/pipeline --branch stop-after -q {d}"
+    # FIXME change branch
+    cmd = f"git clone --depth 1 https://github.com/ooni/pipeline --branch clickhouse2 -q {d}"
```
hellais, line 119: Since the clickhouse2 branch has been merged into master and deleted, this needs to be changed in order for the CI tests to run properly.

newapi/ooniapi/measurements.py
FedericoCeratto: Fixed

newapi/tests/conftest.py
FedericoCeratto: Fixed, good spot!

newapi/ooni_api_uploader.py
```diff
@@ -204,10 +208,12 @@ def main():
             jsonl_s3path = (
                 f"raw/{tstamp[:8]}/{tstamp[8:10]}/{cc}/{testname}/{jsonlf.name}"
             )
-            upload_to_s3(s3, bucket_name, postcanf, postcan_s3path)
-            upload_to_s3(s3, bucket_name, jsonlf, jsonl_s3path)
-
-            update_db_table(db_conn, lookup_list, jsonl_s3path)
+            if conf.get("run_mode", "") == "DESTROY_DATA":
```
hellais, line 211: This sounds like a very dangerous mode of operation to have. Should we maybe add some extra guards (like that you must run the tool with some special option) to prevent this happening by mistake?
At the very least, I think we should display a very big warning when the script is invoked, that it's in this mode of operation.

newapi/ooniapi/auth.py
```diff
@@ -71,7 +73,11 @@ def jerror(msg, code=400):
 
 def create_jwt(payload: dict) -> str:
     key = current_app.config["JWT_ENCRYPTION_KEY"]
-    return jwt.encode(payload, key, algorithm="HS256").decode()
+    token = jwt.encode(payload, key, algorithm="HS256")
+    if isinstance(token, bytes):
+        return token.decode()
+    else:
```
hellais, line 79: nitpick, you don't need the else statement here.

newapi/ooniapi/auth.py
```diff
@@ -356,16 +366,23 @@ def user_login():
...
+            VALUES(:account_id, :role)
+            ON CONFLICT (account_id) DO
+            UPDATE SET role = EXCLUDED.role
+        """
+        q = current_app.db_session.execute(query, query_params).rowcount
+        current_app.db_session.commit()
+
+    if current_app.config["USE_CLICKHOUSE"]:
+        query = """INSERT INTO accounts (account_id, role)
+            VALUES(:account_id, :role)"""
```
hellais, line 382: I guess the `accounts` table in clickhouse is configured to be a `ReplacingMergeTree`?

If so, isn't there the risk that after the "update" of a role is executed we end up with duplicate rows and hence the changes to the account role will not be reflected immediately?

Perhaps we can run an `OPTIMIZE` query right after this (though this feels a bit like a hack)?

hellais: Alternatively we could make use the table mutations: https://clickhouse.com/docs/en/sql-reference/statements/alter/update/

newapi/ooniapi/auth.py
```diff
@@ -411,23 +428,35 @@ def set_account_role():
...
-    # TODO return status
-    return q
+    if current_app.config["DATABASE_URI_RO"]:
+        query = "DELETE FROM accounts WHERE account_id = :account_id"
+        q = current_app.db_session.execute(query, query_params).rowcount
+        current_app.db_session.commit()
+
+    if current_app.config["USE_CLICKHOUSE"]:
+        # reset account to "user" role
+        q = "INSERT INTO accounts (account_id, role) VALUES(:account_id, 'role')"
```
hellais, line 441: Why is the value of `role` set to `'role'`?

Also, this isn't really equivalent to deleting the row from the DB. Perhaps we could make use of the DELETE mutation?
https://clickhouse.com/docs/en/sql-reference/statements/alter/delete/

newapi/ooniapi/auth.py
FedericoCeratto: It's just for readability and safety - to stress that we are running either line 78 or 80

newapi/ooniapi/auth.py
FedericoCeratto: No, these are EmbeddedRocksDB tables. A little database embedded in Clickhouse that provides ACID.

newapi/ooni_api_uploader.py
hellais: Yes I see that, but since it's dependent on something which is inside of a file (that you don't see very clearly when you run it from the CLI), there is a the risk for a mistake.

Can we at least add a warning message to stderr/stdout when it's started in this mode of operation?

newapi/ooniapi/auth.py
hellais: I guess this is a matter of code style. I personally tend to prefer dropping else statements when they are not needed as I find it more readable.

Also, why is it that we need to handle cases in which the token might be in bytes or not? From what I see `jwt.encode` returns an `str`, in which cases is it going to be bytes?

newapi/ooniapi/auth.py
```diff
@@ -124,19 +130,29 @@ def wrapper(*args, **kwargs):
...
+                        return jerror("Authentication token expired", 401)
+
+            else:  # pragma: no cover
+                q = current_app.db_session.execute(query, query_params)
+                row = q.fetchone()
+                if row:
+                    threshold = row[0]
+                    iat = datetime.utcfromtimestamp(tok["iat"])
+                    if iat < threshold:
+                        return jerror("Authentication token expired", 401)
```
hellais, line 155: It looks like these two if else branches have a fair amount of duplication.

I would suggest refactoring this as:

```
            if current_app.config["USE_CLICKHOUSE"]:
                row = query_click_one_row(sql.text(query), query_params)
            else:  # pragma: no cover
                q = current_app.db_session.execute(query, query_params)
                row = q.fetchone()
              if row:
                  threshold = row[0]
                  iat = datetime.utcfromtimestamp(tok["iat"])
                  if iat < threshold:
                      return jerror("Authentication token expired", 401)
```

newapi/ooniapi/auth.py
hellais: Is there the possibility of deleting an item in this key value store?

newapi/ooniapi/measurements.py
```diff
@@ -270,7 +323,31 @@ def _fetch_measurement_body_on_disk(report_id, input: str) -> Optional[bytes]:
...
+    """Fetch raw POST from disk, extract msmt
+    This is used only for msmts that have been processed by the fastpath
+    but are not uploaded to S3 yet.
+    YAML msmts not supported: requires implementing normalization here
+    """
+    assert msmt_uid.startswith("20")
+    tstamp, cc, testname, hash_ = msmt_uid.split("_")
+    hour = tstamp[:10]
+    int(hour)
+    spooldir = Path("/var/lib/ooniapi/measurements/incoming/")
```
hellais, line 336: Should this path perhaps be defined in the config file or at least in a top level const variable?

newapi/ooniapi/measurements.py
```diff
@@ -221,6 +232,48 @@ def _fetch_jsonl_measurement_body(report_id, input: str, measurement_uid) -> byt
...
+@metrics.timer("_fetch_jsonl_measurement_body_clickhouse")
+def _fetch_jsonl_measurement_body_clickhouse(
+    report_id, input: str, measurement_uid
+) -> Optional[bytes]:
+    """Fetch jsonl from S3, decompress it, extract msmt"""
+    inp = input or ""  # NULL/None input is stored as ''
+    try:
+        # FIXME temporary hack to test reprocessing
+        query = """SELECT s3path, linenum FROM new_jsonl
+            PREWHERE report_id = :report_id AND input = :inp
```
hellais, line 244: Nice, I wasn't aware of this PREWHERE optmisation clause.

newapi/ooniapi/measurements.py
```diff
@@ -221,6 +232,48 @@ def _fetch_jsonl_measurement_body(report_id, input: str, measurement_uid) -> byt
...
+        metrics.incr("msmt_not_found_in_jsonl")
+        return None
+
+    s3path = lookup["s3path"]
+    linenum = lookup["linenum"]
+    log.debug(f"Fetching file {s3path} from S3")
+    try:
+        return _fetch_jsonl_measurement_body_inner(s3path, linenum)
+    except:  # pragma: no cover
+        log.error(f"Failed to fetch file {s3path} from S3")
```
hellais, line 273: Should we maybe log the exception?

newapi/ooniapi/measurements.py
```diff
@@ -367,6 +495,103 @@ def get_raw_measurement():
...
+        AND fastpath.report_id = :report_id
+        """
+    query_params = dict(input=input_, report_id=report_id)
+    try:
+        rid_t = datetime.strptime(report_id[:8], "%Y%m%d")
+        query_params["begin"] = rid_t - timedelta(days=30)
+        query_params["end"] = rid_t + timedelta(days=30)
+        query += """
+        AND fastpath.measurement_start_time > :begin
+        AND fastpath.measurement_start_time < :end
```
hellais, line 562: It's worth noting that this will render measurements that are uploaded more than 30 days after their measurement upload time inaccessible.

We should probably note this as a comment and investigate if there is something else we can to do optimise this lookup.

newapi/ooniapi/measurements.py
```diff
@@ -367,6 +495,103 @@ def get_raw_measurement():
...
+    except ValueError:
+        pass  # legacy report_id format without date
+    except Exception as e:
+        log.error(e, exc_info=True)
+
+    query += "LIMIT 1"
+    msmt_meta = query_click_one_row(sql.text(query), query_params)
+    if not msmt_meta:
+        return {}  # measurement not found
+    if msmt_meta["probe_asn"] == 0:
```
hellais, line 573: Perhaps we can make this check more explicit for the leakage.

We can check if probe_asn == 0 and report_id.split("_")[3] != 00 and only in this case return empty?

hellais: We should also link to the issue and blog post explaining why we are doing this:
https://ooni.org/post/2020-ooni-probe-asn-incident-report/
https://github.com/ooni/explorer/issues/495

newapi/ooniapi/private.py
```diff
@@ -769,26 +1046,36 @@ def api_private_country_overview():
...
-    r = q.fetchone()
-    # NOTE: we are hardcoding a fixed first_bucket_date
+    if current_app.config["USE_CLICKHOUSE"]:
+        s = """SELECT
+            toDate(MIN(measurement_start_time)) AS first_bucket_date,
+            COUNT() AS measurement_count,
+            COUNT(DISTINCT probe_asn) AS network_count
+            FROM fastpath
+            WHERE probe_cc = :probe_cc
+            AND measurement_start_time > '2016-01-01'
```
hellais, line 1056: Why is a limit on 2016-01-01 being set here?

newapi/tests/integ/README.adoc
```diff
@@ -1,14 +1,27 @@
...
+ * run the API to receive as set of artificial fresh measurements simulating msmt upload
+   * populate the fastpath table with partially deterministic data
+     * this trigger updates to Clickhouse materialized views and allow testing URL prioritization
+
+   * run the measurement uploader to populate the jsonl table
+     * do not write to S3
+
+Multiple runs of --inject-msmts *may* add new data. Tests should be written to pass if one or more runs have been done.
 
-DATABASE_URL=postgresql://readonly@localhost:5433/metadb pytest-3 -s --show-capture=no -f -x
```
hellais, line 11: Should we maybe move these docs into the top level directory (or maybe even in the top level readme directly)?

newapi/ooni_api_uploader.py
FedericoCeratto: The tool runs as a daemon and it's not meant to be ran manually from CLI.

newapi/ooniapi/auth.py
FedericoCeratto: The duplication in done on purpose to leave the PG code untouched avoiding any risk of breaking it and remove it easily later on.

newapi/ooniapi/citizenlab.py
```diff
@@ -750,13 +753,110 @@ def list_url_priorities():
...
+        AND domain = :domain
+        AND url = :url
+        AND priority = :priority
+        """
+        q = current_app.db_session.execute(query, old).rowcount
+        if q < 1:
+            return jerror("Old rule not found", 400)
+
+    if new:  # add new rule
+        query = """INSERT INTO url_priorities
```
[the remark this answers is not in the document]
FedericoCeratto: Implementing reliable deletion using a temporary table has proven to be quite messy. The current implementation will also allow replication in future.

newapi/ooniapi/measurements.py
```diff
@@ -270,7 +323,31 @@ def _fetch_measurement_body_on_disk(report_id, input: str) -> Optional[bytes]:
...
+def _fetch_measurement_body_on_disk_by_msmt_uid(msmt_uid: str) -> Optional[bytes]:
+    """Fetch raw POST from disk, extract msmt
+    This is used only for msmts that have been processed by the fastpath
+    but are not uploaded to S3 yet.
+    YAML msmts not supported: requires implementing normalization here
+    """
+    assert msmt_uid.startswith("20")
+    tstamp, cc, testname, hash_ = msmt_uid.split("_")
+    hour = tstamp[:10]
+    int(hour)
```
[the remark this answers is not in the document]
FedericoCeratto: It's indeed just a check. We don't want to change type to the hour variable.

newapi/ooniapi/measurements.py
FedericoCeratto: We want to restrict the range for performance reasons. Is there anything different we can do?

newapi/ooniapi/measurements.py
FedericoCeratto: Added

newapi/ooni_api_uploader.py
hellais: What about requiring a command line flag from the daemon init script, so it's clear from reading that script that it's going to behave this way.
My concern is just that we might accidentally have this value set in the config file and not notice it (it's in a config file that is distant from the how the tool is started).

Also the log message is printed only after the files have already been deleted, so we should move printing this further up in the call chain, perhaps at the beginning of the main.

newapi/ooniapi/auth.py
hellais: But why is the value of `role` set to `'role'` instead of `'user'`?

newapi/ooniapi/measurements.py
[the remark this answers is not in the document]
hellais: In that case isn't it perhaps better to explicitly try catch it so it's clear that is what is being done here?

newapi/ooniapi/measurements.py
hellais: > We can check if probe_asn == 0 and report_id.split("_")[3] != 00 and only in this case return empty?

This was not addressed

newapi/ooniapi/private.py
FedericoCeratto: Fixed, thanks.

newapi/ooniapi/measurements.py
FedericoCeratto: Updated

newapi/ooniapi/measurements.py
FedericoCeratto: Dynamic scaling removed.

newapi/ooniapi/citizenlab.py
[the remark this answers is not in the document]
hellais: Since we are using the FINAL keyword for the SELECT query (and CollapseMergeTree table) duplicate entries will be merged when the select is run.

newapi/ooniapi/measurements.py
hellais: We discussed that this is not particularly important, because at this point of the code the caller already has the report_id, so the ASN is anyways leaked and there is no need for the additional check.

newapi/ooniapi/auth.py
```diff
@@ -521,25 +556,51 @@ def set_session_expunge():
...
-
+    query_params: Dict[str, Any] = dict(account_id=account_id)
+    if current_app.config["DATABASE_URI_RO"]:
+        log.info("Deleting from PostgreSQL session_expunge")
+        query = "DELETE FROM session_expunge WHERE account_id = :account_id"
+        current_app.db_session.execute(query, query_params)
+        current_app.db_session.commit()
+        return
+
+    elif current_app.config["USE_CLICKHOUSE"]:
```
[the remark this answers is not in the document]
FedericoCeratto: Updated.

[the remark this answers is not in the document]
hellais: This has been updated to use the same pattern

newapi/ooniapi/auth.py
FedericoCeratto: Added a FIXME

newapi/ooniapi/measurements.py
FedericoCeratto: It's used only in that place. If we'll have to support different paths in future we can change it when needed.

newapi/ooniapi/measurements.py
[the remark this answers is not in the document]
FedericoCeratto: Added comment to clarify.

newapi/ooniapi/measurements.py
FedericoCeratto: Added comment. At this time we never encountered measurements that are not accessible due to this filtering (we are using a generous 60 day range).

FedericoCeratto:
Merging as discussed on Slack

PR #274, opened 2021-12-14
Add torsf statistics

FedericoCeratto:
Initial implementation.
Related to https://github.com/ooni/ooni.org/issues/768

### DIFF
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -1198,6 +1198,98 @@ def get_aggregated():
         return jsonify({"v": 0, "error": str(e)})
 
 
+@api_msm_blueprint.route("/v1/torsf_stats")
+@metrics.timer("get_torsf_stats")
+def get_torsf_stats():
+    """Tor Pluggable Transports statistics
+    Average / percentiles / total_count grouped by day
+    Either group-by or filter by probe_cc
+    Returns a format similar to get_aggregated
+    ---
+    parameters:
+      - name: probe_cc
+        in: query
+        type: string
+        description: The two letter country code
+        minLength: 2
+      - name: since
+        in: query
+        type: string
+        description: >-
+          The start date of when measurements were run (ex.
+          "2016-10-20T10:30:00")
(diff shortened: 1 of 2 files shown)


newapi/ooniapi/measurements.py
```diff
@@ -1198,6 +1198,98 @@ def coalsum(name):
...
+    query = query.group_by(column("measurement_start_day"), column("probe_cc"))
+    query = query.order_by(column("measurement_start_day"), column("probe_cc"))
+
+    try:
+        # disable bitmapscan otherwise PG uses the BRIN indexes instead of BTREE
+        current_app.db_session.execute("SET enable_seqscan=false;")
+        q = current_app.db_session.execute(query, query_params)
+        result = []
+        for row in q:
+            row = dict(row)
```
hellais, line 1281: Should something be done to ensure that the returned data has no "holes" in it? That is to say fill in all dates that have no data for them with an empty datum like we do for other endpoints.

newapi/tests/integ/test_torsf_stats.py
```diff
@@ -0,0 +1,47 @@
...
+
+    response = client.get(url)
+    assert response.status_code == 200
+    assert response.is_json
+    return response.json
+
+
+@pytest.mark.skipif(not pytest.proddb, reason="use --proddb to run")
+def test_torsf_stats_cc(client, log):
+    url = "torsf_stats?probe_cc=IT&since=2021-07-28&until=2021-07-29"
```
hellais, line 20: Do we have some metrics where `anomaly_count` is > 0? If so it might be worth adding a sample check for that too.

hellais (approved):
Left two minor comments, but otherwise looks good to go.

newapi/ooniapi/measurements.py
FedericoCeratto: We don't do this in the aggregation API as the amount of processing and data transfer could increase a lot in some cases. We could add it in future and treat it as optional.

PR #276, opened 2022-01-10
Add support for returning probe_asn and probe_ip as part of check-in

hellais:
~TODO~ DONE:
* [x] Figure out how to test the feature
* [x] Figure out how we want to ship database updates and ensure they are fresh

For the second point I think we either:
* Add the download and ensuring that they are fresh as part of the flask app 
* Add logic for downloading and periodically checking for updates inside of the debian package

Since it's pretty critical that we keep the DBs fresh, I think this should be automated as much as possible


newapi/ooniapi/probe_services.py
```diff
@@ -41,6 +43,27 @@ def generate_report_id(test_name, cc: str, asn_i: int) -> str:
     return rid
 
 
+def get_probe_ip() -> str:
```
FedericoCeratto, line 46: Nitpick "ipaddr" instead of "ip"

newapi/ooniapi/probe_services.py
```diff
@@ -41,6 +43,27 @@ def generate_report_id(test_name, cc: str, asn_i: int) -> str:
...
+    real_ip_headers = [
+        "X-Forwarded-For",
+        "X-Real-IP"
+    ]
+    for h in real_ip_headers:
```
FedericoCeratto, line 51: Maybe it's worth unrolling the loop for only 2 entries

newapi/ooniapi/probe_services.py
```diff
@@ -142,6 +165,17 @@ def check_in():
     run_type = data.get("run_type", "timed")
     charging = data.get("charging", True)
 
+    resp = dict(
+        v=1
+    )
+
+    if probe_cc == "ZZ" and asn == "AS0":
+        probe_ip = get_probe_ip()
+        probe_cc = lookup_probe_cc(probe_ip)
```
FedericoCeratto, line 174: We should handle exceptions here probably. Also what do we want to do for unknown ipaddrs/cc?

newapi/ooniapi/probe_services.py
```diff
@@ -142,6 +165,17 @@ def check_in():
     run_type = data.get("run_type", "timed")
     charging = data.get("charging", True)
 
+    resp = dict(
+        v=1
+    )
+
+    if probe_cc == "ZZ" and asn == "AS0":
```
FedericoCeratto, line 172: Do we want to return CC and ASN in every case instead?

newapi/ooniapi/probe_services.py
```diff
@@ -41,6 +43,27 @@ def generate_report_id(test_name, cc: str, asn_i: int) -> str:
...
+        "X-Real-IP"
+    ]
+    for h in real_ip_headers:
+        if h in request.headers:
+            return request.headers.getlist(h)[0].rpartition(' ')[-1]
+
+    return request.remote_addr
+
+def lookup_probe_asn(ip: str) -> str:
+    with geoip2.database.Reader(current_app.config["GEOIP_ASN_DB"]) as reader:
```
FedericoCeratto, line 58: Do we know how long it takes to open and close the databases every time? Perhaps they should be kept open instead.

newapi/ooniapi/probe_services.py
hellais: I was originally thinking about that, however since probe_cc and probe_asn are user supplied input I think we should ensure that validation of it is very strict.
In order to avoid that I figured we would return it only in cases where it's needed, as otherwise it's basically just echoing back the same data the user supplied.

Or is your suggestion to always do the lookup, even when the user provides probe_cc and probe_asn? In that case, which value should have priority when performing the test list lookup?

newapi/ooniapi/probe_services.py
hellais: That's what I had originally done, but then I figured we might need to eventually enrich this list to include more header values to support more caching proxies. I haven't, for example, tested what headers are used by cloudfront or other caching proxies we might have interest in using like haproxy.

newapi/ooniapi/probe_services.py
hellais: > Also what do we want to do for unknown ipaddrs/cc?

Yeah this is a good point. I guess that for those cases we should, as usual, default to `ZZ` and `AS0` as we usually do, but we might want to also additionally either log this somehow or return the fact that the lookup call failed to the caller via the API response.

Logging is a bit dangerous and tricky, because we don't want to be logging IP addresses of users, but at the same time we would collect some amount of information that will allow us to investigate why the lookup failed so that we can improve our databases.

What are your thoughts?

> Also, should we build a database that returns both CC and ASN in one lookup?

That would be ideal, however we currently have in place only code for generating the ASN database ourselves and we don't have a good strategy for building an IP to CC database since we use the one from https://db-ip.com/.

I think we can start off with this approach and document switching to a single database for future work.

newapi/ooniapi/probe_services.py
hellais: Yeah according to the docs of geoip2 it's recommended to keep a long lived handle on the open database as the process of loading it via the reader is quite IO heavy.

I was originally planning to store these handles in the `g` global flask state, but then realised that it's [only preserved across requests](https://flask.palletsprojects.com/en/2.0.x/api/#flask.g), so it's not really going to help much. Do you have any ideas on how to store some long lived reference to these handles that is local to each flask thread/process?

hellais: Ok nvm, I figured this out. It seems like I just have to attach to the app context which is shared across requests, but thread local.

hellais:
This doesn't need to be deployed to production, it can be deployed just to ams-pg-test.

newapi/ooniapi/probe_services.py
```diff
@@ -41,6 +41,25 @@ def generate_report_id(test_name, cc: str, asn_i: int) -> str:
     return rid
 
 
+def get_probe_ipaddr() -> str:
```
FedericoCeratto, line 44: s/get/extract/

FedericoCeratto (changes requested):
Can we add a comment to clarify that this is WIP code not to be deployed on prod?

newapi/ooniapi/probe_services.py
FedericoCeratto: As an initial iteration we can log the lookup failure and generate a metric without logging the ipaddr.

In future, if the license allows it and if the benefit is worth the effort, we can preprocess the db-ip.com database for faster lookup.

Can you handle lookups leading to unknown / None in this PR?

FedericoCeratto (changes requested):
Added comments

Dockerfile
```diff
@@ -3,10 +3,23 @@ ENV PYTHONUNBUFFERED 1
...
+
+RUN rm -rf /scripts
+
 # Copy code and conf into the container
 COPY newapi /app/
-COPY newapi/api.conf.example /etc/ooni/api.conf
 # Set our work directory to our app directory
 WORKDIR /app/
-# Run runner setup
-RUN ./build_runner.sh
```
FedericoCeratto, line 12: Why is this being removed?

FedericoCeratto (changes requested):

hellais:
Thanks for your comments, I have pushed a new commit addressing them.

newapi/debian/etc/nginx/sites-available/ooni-api.conf
```diff
@@ -1,10 +1,25 @@
 # deployed by ooni-api .deb package
-log_format ooniapi_fmt '"$request" $status';
+
```
FedericoCeratto, line 2: We are deploying this file using ansible. Please remove it from here.

newapi/debian/etc/ooni/api.conf
```diff
@@ -44,5 +44,17 @@ GITHUB_TOKEN = "CHANGEME"
 GITHUB_USER = "CHANGEME"
```
FedericoCeratto, line 44: We are deploying this file using ansible. Please remove it from here.

newapi/debian/ooni_download_geoip.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
FedericoCeratto, line 1: This should generate logs and StatsD metrics so that we can alarm on it.
Also test the database at the end of the download.

newapi/tests/integ/test_probe_services_nodb.py
```diff
@@ -230,6 +230,18 @@ def test_check_in_url_prioritization_category_codes(client, mocks):
     assert len(set(r["url"] for r in c["results"])) == 100
 
 
+def test_check_in_geoip(client, mocks):
```
FedericoCeratto, line 233: This should go in test_probe_services.py

FedericoCeratto (changes requested):

newapi/debian/etc/nginx/sites-available/ooni-api.conf
hellais: This was already part of the debian package. I just updated it so that it would work with the latest version so I could test deploy it.

newapi/debian/ooni_download_geoip.sh
hellais: Done

newapi/debian/ooni_download_geoip.sh
```diff
@@ -0,0 +1,51 @@
...
+mkdir -p /var/lib/ooniapi/
+
+echo "Updating GeoIP database from $CC_URL $ASN_URL" | systemd-cat -t ooniapi-geoip
+curl -o /var/lib/ooniapi/asn.mmdb.gz ${ASN_URL}
+gunzip /var/lib/ooniapi/asn.mmdb.gz
+curl -o /var/lib/ooniapi/cc.mmdb.gz ${CC_URL}
+gunzip /var/lib/ooniapi/cc.mmdb.gz
+
+# Basic smoke test of the geoip databases
+python <<EOF
```
FedericoCeratto, line 38: We should rather switch the rest of the script to Python so it's all the same language. As an added benefit we get to remove the dependency on curl.

FedericoCeratto (changes requested):

newapi/debian/ooni_download_geoip.sh
hellais: Ok done

newapi/debian/ooni_download_geoip.py
```diff
@@ -0,0 +1,119 @@
...
+TS = datetime.utcnow().strftime("%Y-%m")
+ASN_URL = f"https://download.db-ip.com/free/dbip-asn-lite-{TS}.mmdb.gz"
+CC_URL = f"https://download.db-ip.com/free/dbip-country-lite-{TS}.mmdb.gz"
+
+OONI_API_DIR = Path("/var/lib/ooniapi/")
+
+metrics = statsd.StatsClient("127.0.0.1", 8125, prefix="ooni_download_geoip")
+log = logging.getLogger("ooni_download_geoip")
+
+try:
```
FedericoCeratto, line 25: The script can reuse metrics and logging helpers from the ooniapi package

FedericoCeratto (approved):
Later on metrics can be added to report the database size and alarm on unexpected size.

PR #277, opened 2022-01-27
Sync state before returning the value of state

hellais:


newapi/tests/integ/test_citizenlab.py
```diff
@@ -43,13 +43,7 @@ def test_no_auth(client):
...
     r = client.get("/api/v1/url-submission/test-list/global")
     assert r.status_code == 200
-    assert r.json[0] == {
-        "url": "https://2600.org/",
-        "category_code": "HACK",
-        "date_added": "2014-04-15",
-        "source": "citizenlab",
-        "notes": "Updated by OONI on 2017-02-14",
-    }
+    assert r.json[0].keys() == {"url", "category_code", "date_added", "source", "notes"}
```
FedericoCeratto, line 46: They should be `sorted()`

FedericoCeratto (changes requested):

newapi/tests/integ/test_citizenlab.py
hellais: It's a comparison between sets, so order does not matter

PR #279, opened 2022-03-24
Switch more networks to new test helper

FedericoCeratto:


bassosimone (approved):
LGTM, thank you!

newapi/ooniapi/probe_services.py
```diff
@@ -461,9 +461,9 @@ def list_test_helpers():
         ],
     }
     probe_ipaddr = request.headers.get("X-Real-Ip", "")
```
hellais, line 463: Where is the `X-Real-Ip` header being set? I don't see it being set by nginx on the `0.th.ooni.org` host

PR #281, opened 2022-04-19
Improve localdev

hellais:
This PR updates the documentation and tooling around setting up a cross platform dev environment for the OONI API.

In trying to setup my development environment for running locally all the OONI API tests, I realised that the documentation was out of date and that we had so many different ways of doing things (how we recommend in the README vs how tests are run by the CI vs what is in the Makefile) that leads to confusion. I therefore decided to take a stab at improving the situation a bit.

Specifically what I did is as follows:
- Restored a `Makefile` based approach for setting up the dev environment with 3 simple commands: `build`, `initdb` and `tests`
- Improve the `docker-compose.yml` definition to avoid duplicating the work already being done inside of pytest for the database fixture setup
- Updated the `build_runner.sh` script to include recently added dependencies that were missing
- Updated the CI scripts so that they use the same commands as those documented in the Readme
- Removed some dead/duplicate code in the top level directory
- Removed from the Readme the many different ways of setting up the dev environment (I think we should document and support one cross platform method, ie. docker+docker-compose, so as to improve developer experience)

The changes I made should not impact anything non-docker related, so the previous workflows should still be possible.


hellais:
Since I enabled running all the supported tests via the CI, there are now some tests that are failing. I think we should either fix the tests or disable them. @FedericoCeratto some help from you on this front would be greatly appreciated!

PR #282, opened 2022-04-20
Make the aggregation API links downloadable

hellais:
* Set the Content-Disposition header for JSON and CSV formats so as to
trigger a file download in browser instead of opening the file in a page
* Set the correct Content-Type for CSV


hellais:
LGTM!

PR #283, opened 2022-04-25
Localdev clickhouse

hellais:
This is the same as: https://github.com/ooni/api/pull/281, but with the clickhouse branche merged into it.


hellais:
This is superseeded by https://github.com/ooni/api/pull/284

PR #285, opened 2022-05-12
Auth logout

FedericoCeratto:
Add logout route, cleanup.

### DIFF
--- a/newapi/ooniapi/auth.py
+++ b/newapi/ooniapi/auth.py
@@ -95,9 +95,7 @@ def set_JWT_cookie(res, token: str) -> None:
     """Set/overwrite the "ooni" cookie in the browser:
     - secure: used only on HTTPS
     - httponly: block javascript in the browser from accessing it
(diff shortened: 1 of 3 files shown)


newapi/ooniapi/auth.py
```diff
@@ -369,6 +367,25 @@ def user_login():
 """
 
 
+@metrics.timer("user_logout")
+@auth_blueprint.route("/api/v1/user_logout", methods=["GET"])
```
hellais, line 371: I think we should make the logout method work only with `POST` requests. Otherwise it's possible to trigger a logout of any user by having them visit a page which loads something like:
```
<img src="https://api.ooni.io/api/v1/user_logout" />
```

The same is not possible if it's using POST, because that requires using `fetch` or `xhr`, which respects SOP.

FedericoCeratto: Updated

hellais (approved):
LGTM

PR #287, opened 2022-05-16
API codebase cleanup

FedericoCeratto:
Remove PostgreSQL-related code, cleanup tests


newapi/ooniapi/app.py
```diff
@@ -151,18 +160,9 @@ def init_app(app, testmode=False):
...
-    # Setup readers for the geoip database
-    # Note: code related to geoip reading is not yet ready to be deployed in
-    # production.
-    try:
-        app.geoip_cc_reader = geoip2.database.Reader(app.config["GEOIP_CC_DB"])
-        app.geoip_asn_reader = geoip2.database.Reader(app.config["GEOIP_ASN_DB"])
-    except Exception as exc:
-        log.error("Failed to load geoip databases at paths",
-                app.config["GEOIP_CC_DB"], app.config["GEOIP_ASN_DB"])
-        raise exc
```
hellais, line 165: This has changed from being a hard failure at start time, to being a soft failure which might go undetected.

I think it's better to fail early on such types of errors, rather than just letting it go through uncaught.

hellais (commented):
Left a comment on some behaviour that changed during the cleanup. Once that is addressed it can be merged.

newapi/ooniapi/app.py
FedericoCeratto: That's by design. We should not fail to [re]start the whole API if some subsystem is having an issue, otherwise a localized malfunction impact the whole API.

hellais: Fair enough. LGTM

hellais:
LGTM 🐳

PR #290, opened 2022-05-24
Add error handler with localization support

FedericoCeratto:


newapi/error_handlers.json
```diff
@@ -0,0 +1,112 @@
...
+    "ooniapi/auth.py:user_register"
+  ],
+  [
+    "InvalidEmailAddress",
+    "Invalid email address",
+    [],
+    "ooniapi/auth.py:user_register"
+  ],
+  [
+    "EmailError",
```
bassosimone, line 33:
```suggestion
    "EmailSendError",
```

?

newapi/error_handlers.json
```diff
@@ -0,0 +1,112 @@
...
+  ],
+  [
+    "InvalidJwtSignature",
+    "Invalid JWT signature",
+    [],
+    "ooniapi/probe_services.py:_check_probe_token"
+  ],
+  [
+    "UnknownError",
+    "Unknown error {exc}",
```
bassosimone, line 88: Does this mean we're sending the exception to the user?

newapi/error_handlers.json
```diff
@@ -0,0 +1,112 @@
...
+    "ooniapi/probe_services.py:_check_probe_token"
+  ],
+  [
+    "MalformedRequest",
+    "Malformed request",
+    [],
+    "ooniapi/probe_services.py:bouncer_net_tests"
+  ],
+  [
+    "JsonExpected",
```
bassosimone, line 101: I see we have `JSONExpected` and `JsonExpected` and I find it confusing that we're supporting two errors that more or less have the same semantics (I assume) and only differ by capitalization.

newapi/ooniapi/utils.py
```diff
@@ -22,3 +23,23 @@ def nocachejson(*a, **kw):
...
+
+def jer(error_id, tpl, *a, **kw) -> Response:
+    """Error handler with localization support
+    Return error_id, error_msg human friendly msg and errdata dict
+    Set Cache-Control: No-Cache
+    Set HTTP error code from the first optional argument or default to 400
+    """
+    # The function definition is parsed by ../scan_error_handlers.py
+    if error_id == "":
+        error_id = tpl.title().replace(" ", "")
```
bassosimone, line 36: I don't think generating the ID from the template it's a good idea. It seems very non robust and semantically strange. So I can anticipate someone else who's not familiar with these subtle details changing the template and obtaining, as a result, also a different error ID. So, I suggest not to use this surprising pattern. Rather, it would be better, in my opinion, to explicitly mention the correct error ID each time.

newapi/ooniapi/utils.py
```diff
@@ -22,3 +23,23 @@ def nocachejson(*a, **kw):
...
+    if a:
+        http_code = a[0]
+    else:
+        http_code = 400
```
bassosimone, line 40: I found the previous implementation with `code=CODE` more readable than this implementation where we pass a list and then we magically know that the error code is the first element. I see why we cannot revert easily from the previous implementation because we're passing the error data as `kw`. But still it's a bit "meh".

newapi/ooniapi/utils.py
```diff
@@ -22,3 +23,23 @@ def nocachejson(*a, **kw):
...
+    msg = tpl.format(kw)
+    d = dict(error=error_id, error_msg=msg, errdata=kw)
```
bassosimone, line 42: IIUC here we're passing the error that occurred to the client. If my understanding is correct, then I think it's not a good idea to pass detailed error information to the consumer of a web API.

newapi/scan_error_handlers.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
+"""
+Extract error handlers from ooniapi
+
+jer("error code", "error formatter string", a=1, b=2, ...)
+
+Error out if the handlers are different from error_handlers.json
+"""
```
bassosimone, line 8: It is neat that we're importing `ast` to solve this problem. However, this solution brings with it some extra mounts of complexity, such as parsing the `ast`. I would like to know more about the design of this feature and what other alternative designs that could solve the problem were rejected.

hellais (changes requested):
The main issues I see with this approach are that:

1. We are putting the definition of the error strings and error codes directly when they are used. The consequence of this is that:
     a. you need to use external tooling to create an index of all the possible error messages, which is not so good for readability
     b. you increase the risk of duplicating error codes, when you could have used an existing one. This has already happened in this branch with for example `JSONExpected` vs `JsonExpected` and it just going to happen more as the amount of code increases.
     c. you are running the risk of having inconsistencies between error messages and http status code. In the current implementation it's possible to define `jer("", "Authentication required", 401)` and `jer("", "Authentication required", 400)`, which will have the same error message, but different status codes.
      d. if you need to use the same error message in multiple locations, you need to copy paste the text strings from one place to the other

2. Using ast to parse the code manually seems quite tricky and I think we could use an approach that a bit simpler. I also wasn't able to get the `scan_error_handlers.py` script to run, as I got the following error:
```
Scanning function ooniapi/auth.py:role_required
Traceback (most recent call last):
  File "scan_error_handlers.py", line 83, in <module>
    main()
  File "scan_error_handlers.py", line 73, in main
    extract_err_handlers(pyf, node, err_handlers)
  File "scan_error_handlers.py", line 42, in extract_err_handlers
    for eh in scan(desc, n):
  File "scan_error_handlers.py", line 25, in scan
    error_id = x.args[0].value
AttributeError: 'Str' object has no attribute 'value'
```

My proposal for an alternative approach (which is what I started outlining in https://github.com/ooni/api/pull/286/files#diff-01eb929ebf29bd353322afe9fba1921488dc84c3ed03a4d9e9936a8789f1823fR95) is that we have a single module where we define all the HTTP exceptions. These exception objects all have some common properties that we can use to craft well formed messages. We have the assurance we don't make any typos or duplicate strings, because you are importing and returning an objects and not using an opaque string.

It's also going to be easier to write an extractor for the error strings and definitions, because all you have to do is import that module and list all class definitions which are a subclass of the base exception.

This is also the pattern that flask documents or suggest for this type of use-case: https://flask.palletsprojects.com/en/2.1.x/errorhandling/#returning-api-errors-as-json.

PR #291, opened 2022-05-26
Add robots.txt and CSS, cleanup

FedericoCeratto:


newapi/ooniapi/pages/__init__.py
```diff
@@ -40,13 +40,25 @@ def index():
...
 
+@pages_blueprint.route("/css/master.css")
+def serve_master_css() -> Response:
+    """Serve the Bootstrap CSS
+    ---
+    responses:
+      '200':
+        description: CSS
+    """
+    fn = "/usr/lib/python3/dist-packages/ooniapi/static/css/master.css"
```
hellais, line 51: Why are we hardcoding the path to the ooniapi installation? Perhaps we can just serve it via `/static/css/` in the templates

PR #293, opened 2022-06-08
Generate test-list using ASN

FedericoCeratto:
Cleanup and refactoring


newapi/ooniapi/probe_services.py
```diff
@@ -36,9 +36,10 @@ def req_json():
 
 def generate_report_id(test_name, cc: str, asn_i: int) -> str:
     ts = datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
-    cid = "1"  # collector id  TODO read from conf
+    cid = "2"  # collector id  TODO read from conf
```
hellais, line 39: Why was the collector ID changed from 1 to 2?

hellais (approved):
I think we should document why we have changed the collector ID in the generation of the report_id, but beyond that this LGMT

PR #294, opened 2022-06-16
Use new test helpers for all traffic

FedericoCeratto:


newapi/ooniapi/probe_services.py
```diff
@@ -516,32 +516,22 @@ def list_test_helpers() -> Response:
...
-                "front": "d33d1gs9kpq1c5.cloudfront.net",
-                "type": "cloudfront",
-            },
-        ]
-    else:
-        metrics.incr("test_helper_old")
+    # Load-balance test helpers deterministically
+    probe_ipaddr = request.headers.get("X-Real-Ip", "0.0.0.0")
+    try:
+        last_oct = int(probe_ipaddr.rsplit(".", 1)[1])
```
bassosimone, line 522: Please, add a TODO comment mentioning that this code will break with IPv6

bassosimone (commented):
Looks good for moving forward. I suggest adding a comment that the current code would break if using IPv6 and we should increasingly be more aware of IPv6 users.

newapi/ooniapi/probe_services.py
FedericoCeratto: The API and probe services in genera do not support IPv6.

PR #296, opened 2022-06-23
Filter prioritization query by week

FedericoCeratto:


newapi/ooniapi/prio.py
```diff
@@ -135,6 +136,7 @@ def fetch_reactive_url_list(cc: str, probe_asn: int):
     SELECT input, SUM(msmt_cnt) AS msmt_cnt
     FROM counters_asn_test_list
     WHERE probe_cc = :cc
+    AND (week IN (toStartOfWeek(now()), toStartOfWeek(now() - interval 1 week)))
```
hellais, line 139: Doesn’t this mean that we are counting measurements in a range that varies depending on the day of the week? For example if we are at the beginning of the week, it will be ~ measurements from the last week, but if we are towards the end of the week, it will be the measurement count from the last two weeks.

At the very least, if we do this, we should update the docstring of the function to state that it’s not doing “Select measurements count from the last 7 days”, but rather “Select measurements count from the last 7-14 days”

FedericoCeratto: It's using the current and previous week. Doc updated accordingly.

hellais (approved):
LGTM

PR #297, opened 2022-06-27
Reprocessing and switch to Clickhouse

FedericoCeratto:
Move to Clickhouse, support configurable collector ID, multiple bugfixes, cleanup and type checking.

### DIFF
--- a/newapi/ooni_api_uploader.py
+++ b/newapi/ooni_api_uploader.py
@@ -155,9 +155,10 @@ def main():
     bucket_name = conf.get("bucket_name")
     spooldir = Path(conf.get("msmt_spool_dir"))
     format_char = "n"
-    collector_id = conf.get("collector_id", "0")
+    collector_id = conf.get("collector_id")
     identity = f"{format_char}{collector_id}"
     log.info(f"Uploader {collector_id} starting")
+    assert collector_id, "collector_id is not set"
     log.info(f"Using bucket {bucket_name} and spool {spooldir}")
 
     s3 = create_s3_client(conf)
--- a/newapi/ooni_api_uploader.py
+++ b/newapi/ooni_api_uploader.py
@@ -6,7 +6,9 @@ Reads /etc/ooni/api.conf
 
 from configparser import ConfigParser
 from pathlib import Path
+from pathlib import PosixPath as PP
 from datetime import datetime, timedelta
+from typing import List, Dict
 import gzip
 import logging
 import tarfile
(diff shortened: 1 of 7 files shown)


newapi/ooniapi/measurements.py
```diff
@@ -104,7 +102,35 @@ def get_measurement(measurement_id, download=None) -> Response:  # pragma: no co
...
+        body = _fetch_jsonl_measurement_body_from_s3(s3path, linenum)
+    except:  # pragma: no cover
+        log.error(f"Failed to fetch file {s3path} from S3")
+        return make_response("Incorrect or inexistent measurement_id", 400)
+
+    resp = make_response(body)
+    resp.mimetype = "application/json"
+    resp.cache_control.max_age = 3600
+    if download:
+        set_dload(resp, "measurement.json")
```
hellais, line 131: Can we put the measurement_uid inside of the filename here?

newapi/ooni_api_uploader.py
```diff
@@ -95,7 +86,7 @@ def fill_postcan(hourdir, postcanf):
     measurements = []
     postcan_byte_thresh = 20 * 1000 * 1000
     # Open postcan
-    with tarfile.open(str(postcanf), "w") as tar:
+    with tarfile.open(str(postcanf), "w:gz") as tar:
```
hellais, line 89: What has been done with the tarfile that were previously not compressed? Have they been recompressed or do we handle when reading them the fact they might be gzipped or not? If the later, how do we distinguish between the two?

FedericoCeratto: This PR changes only the uploader and this codebase has no tools to reprocess old tarballs or decompress them.
The fastpath should autodetect compression.

newapi/ooniapi/measurements.py
```diff
@@ -80,74 +82,69 @@ def list_files() -> Response:
...
+        body = _fetch_jsonl_measurement_body_from_s3(s3path, linenum)
+    except:  # pragma: no cover
+        log.error(f"Failed to fetch file {s3path} from S3")
+        return make_response("Incorrect or inexistent measurement_id", 400)
+
+    resp = make_response(body)
+    resp.mimetype = "application/json"
+    resp.cache_control.max_age = 3600
+    if download:
+        set_dload(resp, f"{measurement_id}.json")
```
hellais, line 131:
```suggestion
        set_dload(resp, f"ooni_measurement-{measurement_id}.json")
```

newapi/ooniapi/measurements.py
FedericoCeratto: Done

hellais: Can you integrate the suggested change I made below to the filename format?

hellais (approved):
LGTM

PR #299, opened 2022-08-10
Improve reliability

FedericoCeratto:
- [ ] Add database query retry

### DIFF
--- a/newapi/ooniapi/cli/__init__.py
+++ b/newapi/ooniapi/cli/__init__.py
@@ -27,6 +27,7 @@ def shell():
         code.interact(banner="", local=current_app.make_shell_context())
 
 
+# FIXME unused / dead
 @cli.command()
 @with_appcontext
 def create_tables():
--- a/newapi/ooniapi/database.py
+++ b/newapi/ooniapi/database.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 
 from typing import Optional, List, Dict, Union
 import os
+import time
 
 from flask import current_app
 
@@ -14,6 +15,7 @@ from sqlalchemy.sql.selectable import Select
 
 # debdeps: python3-clickhouse-driver
 from clickhouse_driver import Client as Clickhouse
+from clickhouse_driver.errors import NetworkError
 
 
 def _gen_application_name():  # pragma: no cover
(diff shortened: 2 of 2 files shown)


newapi/ooniapi/prio.py
```diff
@@ -37,17 +37,33 @@
...
-    log.info("Started failover_fetch_citizenlab_data")
+    Used only once at startime for fallback."""
+    global fallback_test_items
+    with app.app_context():
+        log = current_app.logger
+        log.info("initializing fallback test-list")
+        try:
+            fallback_test_items = fallback_fetch_citizenlab_data()
+        except Exception:
+            log.error("fallback test-list init error", exc_info=True)
```
hellais, line 56: Should we perhaps log the metric as zero in this case?

newapi/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ooni-api (1.0.18) unstable; urgency=medium
+
+  * Add database query retry
```
hellais, line 3: I don't see where the query retry is implemented in the provided diff

newapi/ooniapi/cli/__init__.py
```diff
@@ -27,6 +27,7 @@ def shell():
         code.interact(banner="", local=current_app.make_shell_context())
 
 
+# FIXME unused / dead
```
hellais, line 30: I suggest removing it in this PR, since you have removed below `oonidata.database.Base`, hence calling this will result in an import error

newapi/ooniapi/probe_services.py
```diff
@@ -244,10 +242,9 @@ def check_in() -> Response:
         )
     except Exception as e:
         log.error(e, exc_info=True)
-        # TODO: use same failover as prio.py:list_test_urls
-        # failover_generate_test_list runs without any database interaction
-        # test_items = failover_generate_test_list(country_code, category_codes, limit)
-        test_items = []
+        # TODO: use same fallback as prio.py:list_test_urls
```
hellais, line 245: Isn't this TODO now implemented?

newapi/ooniapi/prio.py
FedericoCeratto: Incremental counters would not change with a value of 0 although they might record the event perhaps. I'm not sure how that would be useful.

newapi/debian/changelog
FedericoCeratto: That was related to the original change.

hellais: Right, so the changelog should change, no?

newapi/ooniapi/prio.py
FedericoCeratto: I was referring to the metrics.incr() call. The test list size is indeed a gauge and the value will be 0 if the test list is empty.

newapi/debian/changelog
FedericoCeratto: Updated

newapi/ooniapi/cli/__init__.py
FedericoCeratto: Done

newapi/ooniapi/probe_services.py
FedericoCeratto: Removed

PR #300, opened 2022-08-11
Implement support for returning metadata about a domain

hellais:
The category code is the one of the shortest URL in the test list (with
priority to the URL in the global list)

The canonical URL is that of whatever is inside of the global or if not
present there, then it's the one of the shortest variation of the domain

### DIFF
--- a/newapi/ooniapi/private.py
+++ b/newapi/ooniapi/private.py
@@ -844,6 +844,29 @@ def api_private_circumvention_runtime_stats() -> Response:
 
 @api_private_blueprint.route("/domain_metadata")
 def api_private_domain_metadata() -> Response:
+    """
+    The goal of this endpoint is to return the primary category code of a
+    certain domain_name and it's canonical representation.
+    We consider the primary category code to be the category code of whatever is
+    the shortest URL in the test lists giving higher priority to what is in the
+    global list (ex. if we have https://twitter.com/amnesty categorised as HUMR
+    and https://twitter.com/ as GRP, we will be picking GRP as the canonical
+    category code).
+    The canonical representation of a domain is whatever is present in the
+    global list or if it's not present, then the shortest possible
+    representation. 
+    Some research related to this problem was done here:
+    https://gist.github.com/hellais/fab319ae20b0ccca7b548a060ed66e14, where some
+    notes were taken on what fixes need to be done in the test-lists to ensure
+    all of this works as expected (ex. moving shortest URL representations from
+    the country lists into the global list).
+
+    Returns:
+    {
+        "category_code": "CITIZENLAB_CATEGORY_CODE",
+        "canonical_domain": "canonical.tld"
+    }
+    """
     domain = request.args.get("domain")
     if domain is None:
         raise BadRequest("missing domain")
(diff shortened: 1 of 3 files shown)


hellais:
This fixes: https://github.com/ooni/backend/issues/609

newapi/ooniapi/private.py
```diff
@@ -843,3 +841,55 @@ def api_private_circumvention_runtime_stats() -> Response:
 
     except Exception as e:
         return jsonify({"v": 0, "error": str(e)})
+
+
+@api_private_blueprint.route("/domain_metadata")
+def api_private_domain_metadata() -> Response:
+    domain = request.args.get("domain")
```
bassosimone, line 848: Please, include an explanatory message to this algorithm where:

1. you explain the problem we're trying to solve in terms of (a) canonical category of a domain; (b) managing www (the ideal here would link to the issue we're solving here, which I don't know which is)

2. you refer to the your gist where you investigated the test list

newapi/ooniapi/private.py
```diff
@@ -843,3 +841,55 @@ def api_private_circumvention_runtime_stats() -> Response:
...
+        domain_variations.append(domain.lstrip("www."))
+    else:
+        domain_variations.append("www." + domain)
+
+    category_code = "MISC"
+    canonical_domain = sorted(domain_variations, key=len)[0]
+
+    params = {str(idx): d for idx, d in enumerate(domain_variations)}
+    domain_q = "OR ".join([f"domain = %({idx})s " for idx in params.keys()])
+    res = query_click_one_row(
```
bassosimone, line 863:
```suggestion
    # case 1: domain with or without www is in the global list (cc = 'ZZ')
    res = query_click_one_row(
```

newapi/ooniapi/private.py
```diff
@@ -843,3 +841,55 @@ def api_private_circumvention_runtime_stats() -> Response:
...
+        FROM citizenlab 
+        WHERE ({domain_q})
+        AND cc = 'ZZ' 
+        ORDER BY length(url)
+        LIMIT 1
+    """,
+        params,
+    )
+    if not res:
+        res2 = query_click_one_row(
```
bassosimone, line 875:
```suggestion
        # case 2: domain only inside a country list, so we just select the
        # shortest domain among the one with and without www
        res2 = query_click_one_row(
```

bassosimone (approved):
We discussed this change extensively. The proposed solution addresses the two original problems with domain-centric pages (which however still need to be documented in the code before merging, so we know why we added this function when doing future code review and we have a reference to the whole problem).

I do not see any performance issue or impact with respect to the Clickhouse backend given that we're querying a table that is very small by Clickhouse standards.

I do see however issues in the test lists becoming inconsistent unless we apply a CI check that ensures someone does not add https://foobar.com to a country list when https://foobar.com/antani is already in the test list. I think we should at the very minimum create an issue about this topic before merging.

To summarize, please before merging make sure that:

1. we have better in-code documentation
2. we have an issue (or a fix) regarding CI checks in the test lists

Thank you for moving this forward!

🐳

newapi/ooniapi/private.py
hellais: Added more doc strings and a link to the gist

hellais:
Thanks @FedericoCeratto and @bassosimone for the review and input.

I documented what needs to be done to address the issues at the citizenlab test-lists level: https://github.com/ooni/backend/issues/611. I don't think we need to do have this be done prior to merge and deploy. The consequence is just that some categories will be for some amount of less precise.
 
For me we can proceed with merging and deploying this.

PR #301, opened 2022-08-17
Add more strict filtering on inputs on API calls

FedericoCeratto:
Also enable response caching.
Mitigate the impact of vulnerability scanners hitting the API.

### DIFF
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -328,7 +328,7 @@ def get_raw_measurement() -> Response:
     """
     # This is used by Explorer to let users download msmts
     param = request.args.get
-    report_id = param("report_id")
+    report_id = param_report_id()
     if not report_id or len(report_id) < 15:
         raise BadRequest("Invalid report_id")
     input_ = param("input")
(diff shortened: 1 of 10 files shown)


newapi/ooniapi/measurements.py
```diff
@@ -1056,6 +1049,37 @@ def param_url(name):
...
+        raise BadRequest(f"Invalid report_id field")
+    accepted = string.ascii_letters + string.digits + "_"
+    validate(p, accepted)
+    return p
+
+
+def param_report_id() -> str:
+    rid = param_report_id_or_none()
+    if rid is None:
+        raise BadRequest(f"Invalid report_id")
```
ainghazal, line 1066: I would consider returning a less informative err message (maybe an error code?).

FedericoCeratto:
1.0.19~pr301-436 deployed

newapi/ooniapi/measurements.py
FedericoCeratto: It is bubbling up by design to be able to show what field is incorrect. I don't think adding opacity here would be useful because the source code is public.

bassosimone (approved):
🐳

PR #302, opened 2022-08-19
Implement measurement feedback

FedericoCeratto:
Initial MVP. Registered users are allowed to submit, view and edit one feedback item for each msmt.

### DIFF
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -8,7 +8,7 @@ from datetime import datetime, timedelta
 from dateutil.parser import parse as parse_date
 from io import StringIO
 from pathlib import Path
-from typing import Optional, List, Any
+from typing import Optional, List, Any, Dict
 from urllib.parse import urlparse
 import gzip
 import json
@@ -1556,7 +1556,7 @@ def get_msmt_feedback(measurement_uid) -> Response:
     """
     qp = dict(aid=account_id, muid=measurement_uid)
     rows = query_click(sql.text(query), qp)
-    out = dict(summary={})
+    out: Dict[str, Any] = dict(summary={})
     for row in rows:
         status = row["status"]
         if row["is_mine"]:
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -33,6 +33,7 @@ from psycopg2.extensions import QueryCanceledError  # debdeps: python3-psycopg2
 from urllib.request import urlopen
 from urllib.parse import urljoin, urlencode
 
+from ooniapi.auth import role_required, get_account_id_or_none
 from ooniapi.config import metrics
 from ooniapi.utils import cachedjson, nocachejson, jerror
 from ooniapi.models import TEST_NAMES
@@ -1068,6 +1069,13 @@ def param_input_or_none() -> Optional[str]:
     return p
 
 
+def param_measurement_uid() -> str:
+    p = request.args.get("measurement_uid")
+    if not p or len(p) < 10 or len(p) > 100:
+        raise BadRequest(f"Invalid measurement_uid field")
+    return p
+
+
 @api_msm_blueprint.route("/v1/aggregation")
 @metrics.timer("get_aggregated")
 def get_aggregated() -> Response:
@@ -1487,7 +1495,6 @@ def get_torsf_stats() -> Response:
 
 # # measurement feedback
 
-from ooniapi.auth import role_required, get_account_id, jerror
 from ooniapi.database import insert_click
 
 
@@ -1497,7 +1504,6 @@ CREATE TABLE msmt_feedback
     `measurement_uid` String,
     `account_id` String,
     `status` String,
-    `comment` String,
     `update_time` DateTime64(3) MATERIALIZED now64()
 )
 ENGINE = ReplacingMergeTree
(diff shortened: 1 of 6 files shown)


newapi/ooniapi/measurements.py
```diff
@@ -1473,3 +1497,133 @@ def get_torsf_stats() -> Response:
...
+    "ok",
+    "ok.unreachable",
+    "ok.broken",
+    "ok.parked",
+]
+
+
+@api_msm_blueprint.route("/v1/measurement_feedback")
+@metrics.timer("get_msmt_feedback")
+@role_required(["admin", "user"])
```
hellais, line 1543: This should not require authentication as it should be a public API endpoint for any visitor of OONI Explorer.

newapi/ooniapi/measurements.py
```diff
@@ -1473,3 +1497,133 @@ def get_torsf_stats() -> Response:
...
+    LIMIT 1"""
+    qp = dict(account_id=account_id, measurement_uid=measurement_uid)
+    row = query_click_one_row(sql.text(query), qp)
+    if row:
+        return cachedjson("0s", **row)
+
+    return cachedjson("0s")
+
+
+@api_msm_blueprint.route("/v1/submit_measurement_feedback", methods=["POST"])
```
hellais, line 1583: I would suggest placing this under the same path as the `get_msmt_feedback` (i.e. `/measurement_feedback`), but with the POST method instead of get.

newapi/ooniapi/measurements.py
```diff
@@ -1473,3 +1497,133 @@ def get_torsf_stats() -> Response:
...
+    def jparam(name):
+        return request.json.get(name, "").strip()
+
+    log = current_app.logger
+    account_id = get_account_id()
+    measurement_uid = jparam("measurement_uid")
+
+    query = """SELECT status, comment FROM msmt_feedback FINAL
+    WHERE account_id = :account_id AND measurement_uid = :measurement_uid
+    LIMIT 1"""
```
hellais, line 1574: This endpoint should return a list of all the measurement feedbacks that have been given to a specific measurement_uid

newapi/ooniapi/measurements.py
```diff
@@ -1473,3 +1497,133 @@ def get_torsf_stats() -> Response:
...
+      - application/json
+    consumes:
+      - application/json
+    parameters:
+      - in: body
+        required: true
+        schema:
+          type: object
+          properties:
+            measurement_uid:
```
hellais, line 1557: For consistency with the rest of the API, I would suggest putting this as either a query parameter or part of the path itself (e.g. `/v1/measurement_feedback/:measurement_uid:` or `/v1/measurement_feedback?measurement_uid=XXX`)

newapi/ooniapi/measurements.py
```diff
@@ -1473,3 +1497,133 @@ def get_torsf_stats() -> Response:
...
+              type: string
+              description: Measurement ID
+            status:
+              type: string
+              description: Measurement status
+              minLength: 2
+            comment:
```
hellais, line 1606: Until we have a good use-case and UI for this, I would suggest we drop the comment field from it.

newapi/ooniapi/measurements.py
FedericoCeratto: A dedicated entry point is slightly more descriptive "submit_" and easier to search in logs.

newapi/ooniapi/measurements.py
FedericoCeratto: Discussed offline, updated.

newapi/ooniapi/measurements.py
FedericoCeratto: Updated

newapi/ooniapi/measurements.py
FedericoCeratto: Updated

newapi/ooniapi/measurements.py
FedericoCeratto: Updated

FedericoCeratto:
Rebased on master, version bumped to 1.0.25

newapi/ooniapi/measurements.py
hellais: I still see the comment field in the current branch, both in the docstring and the code implementing it.

newapi/ooniapi/measurements.py
hellais: It's considered an anti-pattern to have verbs inside of API endpoints, see:
* https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/#h-use-nouns-instead-of-verbs-in-endpoint-paths
* https://swagger.io/resources/articles/best-practices-in-api-design/#describe-resource-functionality-with-http-methods-4
* https://hevodata.com/learn/rest-api-best-practices/#6
etc.

newapi/ooniapi/measurements.py
FedericoCeratto: Oops, fixed!

newapi/ooniapi/measurements.py
FedericoCeratto: Updated

hellais (approved):
LGTM

FedericoCeratto:
Prod table created

PR #307, opened 2022-09-08
Add LMDB based rate limiting

FedericoCeratto:
Also cleanup some dead code.


FedericoCeratto:
Rebased on master after "Implement measurement feedback #302"

FedericoCeratto:
Rebased after #308 was merged

newapi/ooniapi/rate_limit_quotas.py
```diff
@@ -59,78 +112,64 @@ def __init__(
...
-
-    def _generate_bucket_stats(self):
-        periods = ("month", "week", "day")
-        for b, period in zip(self._ipaddr_buckets, periods):
-            size = len(b)
-            metrics.gauge(f"rate-limit-ipaddrs-{period}", size)
-            print(size)
-
-    def refresh_quota_counters_if_needed(self):
+        # the value is stored as interger in milliseconds
```
hellais, line 123:
```suggestion
        # the value is stored as integer in milliseconds
```

newapi/ooniapi/rate_limit_quotas.py
```diff
@@ -19,14 +19,65 @@
...
+StrBytes = Union[bytes, str]
+
+
+def ipa_lmdb(ipaddr: IpAddress) -> bytes:
+    return ipaddr.packed
+
+
+class LMDB:
+    def __init__(self, dbnames: tuple):
+        self._env = lmdb.open(LMDB_DIR, metasync=False, max_dbs=10)
```
hellais, line 44: I think having the directory hardcoded here might pose some challenges WRT to running tests in different environments as it requires setup of these dirs.

newapi/ooniapi/rate_limit_quotas.py
```diff
@@ -59,78 +112,64 @@ def __init__(
...
+        for hours, limit, db in iterable:
+            vdelta = 1000 * tdelta * limit / hours / 3600
+            with self._lmdb._env.begin(db=db, write=True) as txn:
+                i = txn.cursor().iternext()
+                for raw_ipa, raw_val in i:
+                    v = int(raw_val) + vdelta
+                    if v >= limit:
+                        txn.pop(raw_ipa)
+                    else:
+                        txn.put(raw_ipa, str(int(v)).encode())
```
hellais, line 134: Not sure it's worth the trouble, but maybe since we are going to be doing this a lot of times and for a lot of clients, we might be better of storing the big endian byte representation of the int, instead of having to encode/decode the string representation of them.

newapi/ooniapi/rate_limit_quotas.py
FedericoCeratto: Fixed

newapi/ooniapi/rate_limit_quotas.py
FedericoCeratto: The tests are run in docker at the moment.

newapi/ooniapi/rate_limit_quotas.py
FedericoCeratto: In terms of computational cost it's negligible compared to the rest of the work the API does. I used a string to be able to access the DB directly for debugging but we could indeed store a raw int.

newapi/ooniapi/rate_limit_quotas.py
```diff
@@ -59,78 +112,64 @@ def __init__(
...
-        else:
+        if not ipaddr:
             raise NotImplementedError()
 
-    def get_minimum_across_quotas(self, ipaddr=None, token=None) -> float:
-        assert ipaddr or token
-        if ipaddr:
-            iterable = zip(self._ipaddr_limits, self._ipaddr_buckets)
-            return min(bucket.get(ipaddr, limit) for limit, bucket in iterable)
+        # TODO handle IPv6?
```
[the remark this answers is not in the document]
FedericoCeratto: "Does this mean that if they speak to us on IPv6, they will be effectively by bypassing the quotas?" -> no, this was done before configuring IPv6.
Updated to support IPv6 and tested on ams-pg-test.

newapi/ooniapi/rate_limit_quotas.py
```diff
@@ -59,78 +114,66 @@ def __init__(
             self._whitelisted_ipaddrs.add(ipaddress.ip_address(ipa))
```
hellais, line 122: It's not part of the newly added code, but based on the type hints for 
```
        unmetered_pages=Optional[List[str]],
```

It can be `null`, yet there is no check for the value being null in the for loop.

hellais (approved):
LGTM

FedericoCeratto:
1.0.28~pr307-488 deployed

PR #309, opened 2022-10-07
Add locking to citizenlab

FedericoCeratto:
Also fix time range filtering in aggregation.


FedericoCeratto:
1.0.29~pr309-498 deployed on FSN

PR #312, opened 2022-10-24
Hide DB exception traceback

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/619

### DIFF
--- a/newapi/ooniapi/database.py
+++ b/newapi/ooniapi/database.py
@@ -1,8 +1,4 @@
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
-from __future__ import unicode_literals
-
+import logging
 from hashlib import shake_128
 from typing import Optional, List, Dict, Union
 import os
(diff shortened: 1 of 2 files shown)


newapi/ooniapi/database.py
```diff
@@ -58,9 +57,17 @@ def _run_query(query: Query, query_params: dict, query_prio=3):
...
+    try:
+        q = current_app.click.execute(
+            query, query_params, with_column_types=True, settings=settings
+        )
+    except clickhouse_driver.errors.ServerException as e:
+        title = e.message.splitlines()[0]
+        if title.startswith("DB::Exception"):
+            log.info(e.message)
+            title = title.split(" ", 1)[1]
+        raise Exception(title)
```
hellais, line 69: Since we are already the full database exception stacktrace to our logger, how about we just return a generic error message stating that there was an error with the query?

That's because here we are making an assumption about the way in which stacktraces are formatted by clickhouse and if that doesn't apply anymore in the future (or even in some cases we haven't tested), we might fail open and return more than we should.

FedericoCeratto: Ok, updated.

hellais (approved):
LGTM

FedericoCeratto:
1.0.32~pr312-514 deployed to FSN

PR #313, opened 2022-10-25
Add Authorization Bearer header

FedericoCeratto:
- [x] Add Authorization Bearer header
- [x] Remove cookie
- [x] Introduce `/api/v1/user_refresh_token` entry point
- [x] Minor refactor

### DIFF
--- a/.github/workflows/test_new_api.yml
+++ b/.github/workflows/test_new_api.yml
@@ -35,6 +35,7 @@ jobs:
       #  run: docker ps -a
 
       - run: find / -name pytest.xml
+        if: success() || failure() # run even if previous step failed
 
       - name: Test report
         uses: dorny/test-reporter@v1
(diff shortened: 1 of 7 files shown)


FedericoCeratto:
Rebased on master

hellais (commented):
Looking at the implementation of the client-side bit of this (https://github.com/ooni/test-lists-ui/pull/68/files), I gather that we are basically doing a token refresh request prior to any userFetcher request. Maybe we can improve this by returning the expiry time of the token and only doing a refresh if the token is close to expiry (< 8h?) and on any request that requires authentication. I believe we need to return this as an extra parameter in the refresh token response and the login response as the client cannot decrypt the JWT token which is encrypted using the server secret.

For the rest the PR looks good to me.

FedericoCeratto:
1.0.33~pr313-526 deployed on prod as discussed

PR #314, opened 2022-11-18
MAT: support time granularity

FedericoCeratto:
https://github.com/ooni/backend/issues/623

### DIFF
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -1169,7 +1169,7 @@ def get_aggregated() -> Response:
       - name: time_gran
         in: query
         type: string
-        description: Time granularity
+        description: Time granularity. Used only when the X or Y axis represent time.
         enum:
         - hour
         - day
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -1174,6 +1174,7 @@ def get_aggregated() -> Response:
         - hour
         - day
         - week
+        - month
         - year
         - auto
       - name: axis_x
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -1252,7 +1252,7 @@ def get_aggregated() -> Response:
     ]
     table = sql.table("fastpath")
     where = []
-    query_params = {}
+    query_params: Dict[str, Any] = {}
 
     if domain:
         where.append(sql.text("domain = :domain"))
@@ -1477,7 +1477,7 @@ def get_torsf_stats() -> Response:
     ]
     table = sql.table("fastpath")
     where = [sql.text("test_name = 'torsf'")]
-    query_params = {}
+    query_params: Dict[str, Any] = {}
 
     if probe_cc:
         where.append(sql.text("probe_cc = :probe_cc"))
--- a/newapi/ooniapi/measurements.py
+++ b/newapi/ooniapi/measurements.py
@@ -1166,7 +1166,7 @@ def get_aggregated() -> Response:
         description: >-
           The end date of when measurement were run (ex.
           "2016-10-20T10:30:00")
-      - name: time_gran
+      - name: time_grain
         in: query
         type: string
         description: Time granularity. Used only when the X or Y axis represent time.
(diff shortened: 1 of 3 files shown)


newapi/ooniapi/measurements.py
```diff
@@ -1166,6 +1166,16 @@ def get_aggregated() -> Response:
...
           "2016-10-20T10:30:00")
+      - name: time_gran
+        in: query
+        type: string
+        description: Time granularity
+        enum:
+        - hour
+        - day
+        - week
+        - year
```
hellais, line 1177: Can we add month here? If it's too many aggregation, I would suggest dropping `year` in favour of `month` as that is a much more common use-case.

newapi/ooniapi/measurements.py
```diff
@@ -1335,16 +1294,37 @@ def _clickhouse_aggregation(
...
-            delta = abs((until - since).days)
+            delta = until - since
         else:
-            delta = 0
+            delta = None
+
+        ranges = (
+            (7, ("hour", "day", "auto")),
+            (30, ("day", "week", "auto")),
+            (365, ("day", "week", "month", "auto")),
```
hellais, line 1306: Ah I now see that here it's using `month`, so maybe the docstring is incorrect.

newapi/ooniapi/measurements.py
```diff
@@ -1335,16 +1294,37 @@ def _clickhouse_aggregation(
...
+            raise Exception("Invalid since and until values")
 
-        # TODO: add a granularity parameter and a check around query weight /
-        # / response size. Also add support in CSV format.
-        granularity = ("day", "toDate")
-        tcol, fun = granularity
+        for thresh, allowed in ranges:
+            if delta > timedelta(days=thresh):
+                continue
+            assert time_gran in allowed
```
hellais, line 1314: It would be nice in this case to return a meaningful error to the end user stating which are the allowed aggregation intervals for the current time window.

newapi/ooniapi/measurements.py
```diff
@@ -1166,6 +1166,16 @@ def get_aggregated() -> Response:
         description: >-
           The end date of when measurement were run (ex.
           "2016-10-20T10:30:00")
+      - name: time_gran
```
hellais, line 1169: I think the name of this argument could be improved. `time_gran` is not particularly understandable and the doc string "Time granularity" doesn't really do much to help explain how it will affect the output data.

Maybe it could be called "tick_every" (which is commonly used in charting applications to indicate the ticks on the axis).

Alternatively it could perhaps simply be called "group_by"?

newapi/ooniapi/measurements.py
```diff
@@ -1361,15 +1341,19 @@ def add_axis(axis, cols, colnames, group_by):
         group_by.append(column(axis))
 
     group_by: List = []
-    if axis_x == "measurement_start_day":
-        group_by_date(since, until, cols, colnames, group_by)
-    elif axis_x:
-        add_axis(axis_x, cols, colnames, group_by)
+    try:
+        if axis_x == "measurement_start_day":
```
hellais, line 1345: Should we perhaps return an error if the user has chosen a `group_by` value, but is not aggregating on the `measurement_start_day` axis?

Also, we probably ought to change the name of `measurement_start_day` to `measurement_start_time`, since it's no longer the `day`, but rather something that changes depending on the value of the `group_by` field.

hellais:
This looks great! Thanks for putting it together. I left some comments for some improvements I think we should make prior to merge.

newapi/ooniapi/measurements.py
FedericoCeratto: We discussed changing the name with @majakomel but it will take a second PR in order not to break production during the transition

newapi/ooniapi/measurements.py
FedericoCeratto: Fixed

newapi/ooniapi/measurements.py
FedericoCeratto: It's described in the "description:" field. It contains "time_" to clarify that it's related specifically to time. I added a better description.

newapi/ooniapi/measurements.py
FedericoCeratto: Added together with a new set of time ranges.

newapi/ooniapi/measurements.py
hellais: We discussed out of band that maybe a name that preserves both being short and understandable is `time_grain`, so let's go for that.

FedericoCeratto: Moved to time_grain

hellais (approved):
LGTM

FedericoCeratto:
@majakomel updated as discussed, 1.0.34~pr314-534 deployed to -test

PR #319, opened 2023-02-10
Fixed keys for user reporting mechanism

agrabeli:
The keys for the user reporting mechanism had the values `unreachable`, `broken` and `parked` nested under `ok` (which pertains to successful measurements), which is not accurate.

This PR removes the `unreachable` value (which generally qualifies under `blocked`, but we already have more specific values under `blocked`), and creates nested values (including `parked`) under `broken`.


hellais (approved):
LGTM

FedericoCeratto:
Related https://github.com/ooni/api/pull/322

